PR #4, opened 2023-03-20
Add RHEL 7 kernel cleanup steps from documentation.

jeffmcutter:


djdanielsson (approved):
can we do this without shell?

djdanielsson (changes requested):
even if you leave the shell stuff you need to fix the lint issues

djdanielsson (approved):
LGTM

PR #5, opened 2023-03-20
Add additional repos for leapp upgrade.

jeffmcutter:
Add code ready and Satellite client repositories so that leapp can find all packages.


djdanielsson (approved):
LGTM

PR #6, opened 2023-03-21
Add grubby check for RHEL 7 to 8 upgrade.

jeffmcutter:
Also, exclude collections/ from linting to be able to commit.


djdanielsson (approved):
LGTM

PR #7, opened 2023-03-21
Add selinux leapp post steps.

jeffmcutter:


roles/upgrade/tasks/leapp-post-upgrade-selinux.yml
```diff
@@ -0,0 +1,27 @@
+---
+- name: Ensure there are no SELinux denials
+  ansible.builtin.shell: >
+    ausearch -m AVC,USER_AVC -ts boot 2>&1
+  register: ausearch_results
+  changed_when: false
+  failed_when: ausearch_results.stdout != "<no matches>"
+
+- name: SELinux mode is set to {{ selinux_mode }}
+  ansible.posix.selinux:
```
djdanielsson, line 10: where are we stating this is a requirement for this collection to work?

djdanielsson (changes requested):
I think we are missing something

djdanielsson (approved):
I think this should be all we need to do, lets verify it soon

roles/upgrade/tasks/leapp-post-upgrade-selinux.yml
swapdisk: Right, this is going to fail the upgrade playbook if there are any denials reported, but how would we know they are because of the upgrade? For example, if there was an app causing benign denials even before the upgrade, we would expect to see the same after the upgrade even if they are not impactful.

PR #9, opened 2023-03-23
Snapmgt dev and proto vcenter.yml

jwesterd-redhat:
simple snapmgt for  LVM snaps and Vcenter.yml  for ESX snapshots


djdanielsson (changes requested):
I don't think the changes to vcenter are really worth it and should be removed

PR #16, opened 2023-03-24
Add Satellite Activation Keys.

jeffmcutter:
Switch from shell to community.general.redhat_subscription.
Add additional registration to enable leapp activation key in analysis.
Variable-ize Satellite Org and Activation Keys.


djdanielsson (approved):
LGTM

PR #19, opened 2023-03-27
General clean up 1.

jeffmcutter:


roles/logging/tasks/main.yml
```diff
@@ -1,33 +1,32 @@
...
+    mode: '0755'
 
-- name: Create new log file
-  ansible.builtin.template:
-    src: templates/logfile.j2
-    dest: "{{ log_file }}"
-    owner: root
-    group: root
-    mode: 0644
-    validate: "test ! -e {{ log_file }} -a %s"
```
djdanielsson, line 18: is this validate not worth keeping?

roles/upgrade/tasks/leapp-post-upgrade.yml
```diff
@@ -8,30 +8,31 @@
...
-        cd /lib/modules && /usr/bin/ls -d *.el7*
+      ansible.builtin.find:
+        paths:
+          - /lib/modules
+        patterns: '*.el7*'
+        file_type: directory
       register: old_kernels
-      changed_when: false
 
     - name: Remove weak modules from old kernels
```
djdanielsson, line 18: I wonder if this should be moved to validation?

djdanielsson (approved):
assuming you do not think there needs to be changes in the areas I have questions and have a reason for how it is, I am good with these changes

PR #22, opened 2023-03-29
Move inode-check to RHEL 6 to 7 analysis.

jeffmcutter:


djdanielsson:
What is the benefit of moving this to its own file?

jeffmcutter:
@swapdisk Indicated that the inode check isn't helpful on the Leapp upgrades but that it would still be useful for the RHEL 6 to 7 upgrades.  This PR removes the inode check from the Leapp analysis and leaves a stub for it for future RHEL 6 to 7 automation.

swapdisk (approved):
LGTM...

PR #24, opened 2023-03-31
Method to snap OS and then leapp upgrade with exclusion and rollback

jwesterd-redhat:
Here is a method to capture snapshots of selected file systems in support of Leapp preupgrade and upgrade methods.    Rollback is tested several times on RHEL7 to RHEL8 upgrade with 2 tested stages.   Rollback from Pre-reboot (ie RHEL7 -> RHEL7 - had run the "leapp upgrade" but want to rollback) and post-reboot (I am finished completely with RHEL7 -> RHEL8 upgrade.   Note: this method stuffs the payload from /var/lib/leapp into /var/tmp/leapp and requires about 5 GB in /var/tmp/leapp.

### DIFF
--- a/roles/snapshot/files/snapmgt.bash
+++ b/roles/snapshot/files/snapmgt.bash
@@ -1,0 +1,277 @@
+#!/bin/bash
+#
+# Script: snapmgt.bash
+#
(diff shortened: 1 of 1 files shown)


djdanielsson:
lets wait to merge this for the moment but leave the PR open

PR #26, opened 2023-04-03
AWS snapshot

heatmiser:
Automation for performing snapshot creation, reverting, and deletion with Amazon EC2 instances and EBS-backed storage.


heatmiser:
Need to redo this PR.

PR #29, opened 2023-04-03
Disable update availability check for now, add TODO.

jeffmcutter:


djdanielsson (changes requested):
I'm not sure I understand the issue here and what this fixes we can discuss later but I don't want to merge yet

PR #31, opened 2023-04-05
Update previous packages check to look for all prior  OS versions.

jeffmcutter:


djdanielsson (approved):
LGTM, it would be nice to figure out how to do this with ansible at some point instead of shell/cmd

PR #33, opened 2023-04-05
Add first boot success check for migration results.

jeffmcutter:


roles/upgrade/tasks/upgrade-validation.yml
```diff
@@ -2,19 +2,21 @@
...
+- name: Determine leapp_dest_version from rhel 8
+  ansible.builtin.set_fact:
+    leapp_dest_version: "{{ rhel_8_upgrade_dest_version }}"
+  when: ansible_facts.ansible_local.pre_ripu.distribution_major_version == '8'
+
+- name: Validate current OS version
   ansible.builtin.assert:
-    that: ansible_distribution_version == rhel_8_upgrade_dest_version
-    fail_msg: Expected leapp destination OS version {{ rhel_8_upgrade_dest_version }} but OS version is {{ ansible_distribution_version }}
+    that: ansible_distribution_version == leapp_dest_version
```
djdanielsson, line 17: do you think we should combine this assert with the other one?

jeffmcutter: I'm not sure I understand the request.  The asserts take half a second to run, and it is clear what they are checking for separately.  Based on that I would say no, we shouldn't combine them.

PR #37, opened 2023-04-10
Convert modprobe to use Ansible module.

jeffmcutter:


djdanielsson (changes requested):
Do we think this is something that many people would want?

roles/analysis/defaults/main.yml
```diff
@@ -47,4 +47,7 @@ leapp_repos_enabled: []
 #   - satellite-client-6-for-rhel-{{ ansible_distribution_major_version | int + 1 }}-x86_64-rpms
 
 result_filename: '/var/log/leapp/leapp-report.txt'
+
+kernel_modules_to_unload:
+  - btrfs
```
djdanielsson, line 52: We don't want any defaults for this if we keep it

roles/analysis/tasks/analysis-leapp.yml
```diff
@@ -56,6 +56,12 @@
     mode: '0644'
   when: leapp_answerfile is defined
 
+- name: Kernel modules defined by kernel_modules_to_unload are unloaded
+  community.general.modprobe:
+    name: "{{ item }}"
+    state: absent
+  loop: "{{ kernel_modules_to_unload }}"
```
djdanielsson, line 63: when: kernel_modules_to_unload | length is not match('0')

djdanielsson (approved):
LGTM

PR #45, opened 2023-04-19
Removing snapshot/rollback bits

swapdisk:
Automation of snapshot/rollback capabilities are out of scope for the ansible-leapp collection. Removing these roles from the here, but we are planning to maintain them elsewhere.


djdanielsson (approved):
LGTM

PR #46, opened 2023-04-27
Remove validation role.

jeffmcutter:
I believe we agreed to remove the validation role on the brain dump two weeks ago.


djdanielsson (approved):
LGTM

PR #53, opened 2023-05-01
Add initial RHEL 6 preupg analysis.

jeffmcutter:


roles/analysis/tasks/main.yml
```diff
@@ -19,12 +19,19 @@
     content: "{{ ansible_facts | ansible.builtin.combine({'ansible_local': {}}) }}"
     dest: /etc/ansible/facts.d/pre_ripu.fact
 
-# TODO: Include tasks for RHEL 6 to 7 non-leapp upgrade
-# - name: Include tasks for preupg assistant analysis
-#   ansible.builtin.include_tasks: analysis-preupg.yml
-#   when: ansible_distribution_major_version|int < 7
+- name: Include tasks for preupg assistant analysis
+  ansible.builtin.include_tasks: analysis-preupg.yml
```
djdanielsson, line 23: rename the file to have rhel6 if it is specifically for rhel 6

djdanielsson (changes requested):

djdanielsson (approved):
LGTM

roles/analysis/handlers/main.yml
```diff
@@ -6,4 +6,10 @@
     activationkey: "{{ satellite_activation_key_pre_leapp }}"
     org_id: "{{ satellite_organization }}"
     force_register: true
+
+- name: Remove lynx package
```
swapdisk, line 10: @jeffmcutter, why are we removing this package during analysis? Ideally, the analysis role shouldn't be making changes beyond installing the packages required to generate the report.

swapdisk: Got it, thanks!

PR #72, opened 2023-06-05
Fix sanity failures

swapdisk:
Let's see if this works to fix sanity test failures?


swapdisk:
Here's what I did: 
- Add quotes as recommended for check-inodes.sh
- Ignore shebang sanity test because of bug ansible/ansible#78346

Now all CI tests get passed.

djdanielsson (approved):
LGTM

PR #76, opened 2023-06-07
Fix leapp cleanup

jeffmcutter:
When upgrading from 7.9 to 8.8, some leapp packages for 7.9 remained causing the subsequent yum update to fail.


swapdisk (approved):
LGTM.

PR #77, opened 2023-06-07
Post upgrade validate major os version

jeffmcutter:
Since this will be a moving target and cause maintenance issues, checking the major version ensures that the upgrade was successful assuming the correct repositories were provided.


djdanielsson (approved):
LGTM

PR #85, opened 2023-06-13
Add task timeout to reboots

jeffmcutter:
Infrequently the reboot module fails to detect that the system has rebooted successfully and is back online.  The reboot_timeout also fails to detect the time out.  Adding the task level timeout should ensure that the job doesn't wait indefinitely.


djdanielsson (approved):
LGTM

PR #87, opened 2023-06-14
Enhance leapp_upgrade_type to support satellite, cdn, and rhui.

jeffmcutter:
Instead of the confusing and less useful connected and disconnected options previously supported.


djdanielsson (approved):
LGTM

PR #88, opened 2023-06-14
Remove pythone2-leapp package with other leapp packages.

jeffmcutter:
7.9 -> 8.8 packages are not being cleaned they way 7.9 -> 8.6 packages were.


djdanielsson (changes requested):
rebase

PR #94, opened 2023-06-22
WIP - Capture and review non-rhel versioned packages 

scott-vick:
roles/analysis
  - Updated main.yml to capture a list of non-rhel versioned packages and save the list for comparison after the upgrade roles/upgrade
  - Updated check-for-old-packages.yml to capture a post-upgrade list of non-rhel versioned packages and compare them to the pre-upgrade list.
    - Prints a warning message and list of missing packages (if any packages were removed during the upgrade).

### DIFF
--- a/roles/upgrade/tasks/check-for-old-packages.yml
+++ b/roles/upgrade/tasks/check-for-old-packages.yml
@@ -8,24 +8,48 @@
       grep -ve '[\.|+]el{{ ansible_distribution_major_version }}' |
       grep -vE '^(gpg-pubkey|libmodulemd|katello-ca-consumer)' |
       sort
-  register: previous_version_packages
+  register: unsigned_packages_post
   changed_when: false
   failed_when:
-    - previous_version_packages.rc != 0
-    - previous_version_packages.stderr != ""
+    - unsigned_packages_post.rc != 0
+    - unsigned_packages_post.stderr != ""
+
+- name: Set a fact that lists the old packages and packages not versioned by rhel release
+  ansible.builtin.set_fact:
+    unsigned_packages_post: "{{ unsigned_packages_post.stdout_lines }}"
+    cacheable: true
+
+- name: Diff the list of current un-versioned packages and the previous list of un-versioned packages and set fact for missing packages
+  ansible.builtin.set_fact:
+    missing_non_rhel_packages: "{{ ansible_facts.ansible_local.non_rhel_packages | difference(unsigned_packages_post) }}"
+
+- name: Display warning message if there are any missing non-rhel versioned packages from the pre-upgrade
+  ansible.builtin.debug:
+    msg:
+      - "Warning!! There are non-rhel packages that were removed during the upgrade. Please review the list of missing packages"
+      - "{{ missing_non_rhel_packages }}"
+  when: missing_non_rhel_packages | length >= 1
+
+- name: Save list of missing packages for future use/review
+  ansible.builtin.copy:
+    content: "{{ missing_non_rhel_packages }}"
+    dest: /etc/ansible/facts.d/missing_packages.fact
+    mode: '0644'
+    owner: root
+    group: root
 
 - name: Display results of search for old packages and packages not versioned by rhel release
   ansible.builtin.debug:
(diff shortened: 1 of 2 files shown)


roles/upgrade/tasks/check-for-old-packages.yml
```diff
@@ -14,18 +14,37 @@
...
+    postupgrade_version_packages: "{{ previous_version_packages.stdout_lines }}"
+    cacheable: yes
+
+- name: Diff the list of current un-versioned packages and the previous list of un-versioned packages and set fact for missing packages
+  ansible.builtin.set_fact:
+    missing_package_list: "{{ ansible_facts.ansible_local.pre_ripu.unsigned_packages_pre | difference(unsigned_packages_post) }}"
+
+- name: Display warning message if there are any missing non-rhel versioned packages from the pre-upgrade 
+  ansible.builtin.debug:
+    msg: "Warning!! There are non-rhel packages that were removed during the upgrade. Please review the list of missing packages"
```
djdanielsson, line 28: Combine both debug messages into one task with multiple lines

djdanielsson (changes requested):

roles/upgrade/tasks/check-for-old-packages.yml
```diff
@@ -14,18 +14,37 @@
     - previous_version_packages.rc != 0
     - previous_version_packages.stderr != ""
 
+- name: Set a fact that lists the old packages and packages not versioned by rhel release
+  ansible.builtin.set_fact:
+    postupgrade_version_packages: "{{ previous_version_packages.stdout_lines }}"
+    cacheable: yes
```
djdanielsson, line 20: This should be true not yes

djdanielsson (changes requested):

roles/upgrade/tasks/check-for-old-packages.yml
scott-vick: Updates completed as requested.

roles/upgrade/tasks/check-for-old-packages.yml
scott-vick: Updates completed as requested.

roles/upgrade/tasks/check-for-old-packages.yml
```diff
@@ -23,13 +23,21 @@
   ansible.builtin.set_fact:
     missing_package_list: "{{ ansible_facts.ansible_local.pre_ripu.unsigned_packages_pre | difference(unsigned_packages_post) }}"
 
-- name: Display warning message if there are any missing non-rhel versioned packages from the pre-upgrade 
+- name: Display warning message if there are any missing non-rhel versioned packages from the pre-upgrade
   ansible.builtin.debug:
-    msg: 
-      - "Warning!! There are non-rhel packages that were removed during the upgrade. Please review the list of missing packages"
+    msg:
+      - "Warning!! There are non-rhel packages that were removed during the upgrade. Please review the list of missingb packages"
```
djdanielsson, line 29: typo, missing not missingb

roles/upgrade/tasks/check-for-old-packages.yml
```diff
@@ -23,13 +23,21 @@
...
       - "{{ missing_package_list }}"
   when: missing_package_list | length >= 1
 
+- name: Save all of the facts from the run for future use/review
+  ansible.builtin.copy:
+    content: "{{ ansible_facts | ansible.builtin.combine({'ansible_local': {}}) }}"
```
djdanielsson, line 35: I don't think we want to save them all just the one fact missing_package_list

djdanielsson (changes requested):
you also need to save the list of original packages to the local facts so it is tracked outside of this play

roles/upgrade/tasks/check-for-old-packages.yml
scott-vick: Updated so it saves as a single local fact called "missing_package_list"

scott-vick:
> you also need to save the list of original packages to the local facts so it is tracked outside of this play

Did that inside the main.yml tasks in the analysis role.

djdanielsson (approved):
LGTM

PR #97, opened 2023-07-03
Analysis-docs

jwesterd-redhat:


jwesterd-redhat:
docs for analysis role

swapdisk (changes requested):
"This is used during planning phase for enterprise IPU plans, as well as during each systems IPU journey."

Planning phase for plans? IPU uses as an adjective twice in the same sentence? 

Awkward grammar aside, I don't know what we are trying to accomplish here. Enterprises are not the only users for this collection. Why should we be addressing the challenges of enterprise change management processes with our role docs?

jwesterd-redhat (commented):
Rewrote the line that was targeted at enterprise method, and left it more amenable to users and enterprise as well.

PR #98, opened 2023-07-05
Change default for update_grub_to_grub_2 from true to false.

jeffmcutter:


djdanielsson (changes requested):
update changelog please

djdanielsson (approved):
LGTM

PR #102, opened 2023-07-06
update analysis README.md

jwesterd-redhat:


roles/analysis/README.md
```diff
@@ -1,27 +1,65 @@
...
+
+In an upcomoing release, the metadata will be included as part of the RPM install of leapp RPMs.
+
+For more information, refer to the knowledge article [Leapp utility metadata in-place upgrades of RHEL for disconnected upgrades (including Satellite)](https://access.redhat.com/articles/3664871).
+
+## Commonly modified variables
+
+| Name                  | Type | Default value           | Description                                     |
+|-----------------------|------|-------------------------|-------------------------------------------------|
+| leapp_upgrade_type    | String  | satellite | satellite, cdn or rhui |
```
swapdisk, line 21: Previous description was better. Restore like this... 

```suggestion
| leapp_upgrade_type    | String  | "satellite" | Set to "cdn" for hosts registered with Red Hat CDN and "rhui" for hosts using rhui repos. |
```

roles/analysis/README.md
```diff
@@ -1,27 +1,65 @@
...
+| leapp_preupg_opt | String | --no-rhsm when leapp_upgrade_type is connected | Upstream repository usage - whether to use RHSM |
```
swapdisk, line 22: Previous description was better. Explanation of the default value is confusing. If we really want to provide a way for users to append extra options to the leapp command, we should use a different variable that won't interfere with our default logic. This is an optional variable, so it should go in your new optional section. 

Move to optional section and change to this...
```suggestion
| leapp_preupg_opt | String |  | Optional string to define command line options to be passed to the `leapp` command when running the pre-upgrade. |
```

roles/analysis/README.md
```diff
@@ -1,27 +1,65 @@
...
 
-## Role variables
 
-| Name                    | Default value         | Description                                         |
-|-------------------------|-----------------------|-----------------------------------------------------|
-| leapp_upgrade_type      | "satellite"           | Set to "cdn" for hosts registered with Red Hat CDN and "rhui" for hosts using rhui repos. |
-| leapp_metadata_url      |                       | See Requirements section above.                     |
-| leapp_answerfile        |                       | Optional multi-line string. If defined, this will be used as the contents of `/var/log/leapp/answerfile`. |
-| leapp_preupg_opts       |                       | Optional string to define command line options to be passed to the `leapp` command when running the pre-upgrade. |
-| post_reboot_delay       | 120                   | Optional integer to pass to the reboot post_reboot_delay option. |
```
swapdisk, line 19: Why did we remove post_reboot_delay variable. Add it back under optional section.

roles/analysis/README.md
```diff
@@ -1,27 +1,65 @@
...
+
+For more information, refer to the knowledge article [Leapp utility metadata in-place upgrades of RHEL for disconnected upgrades (including Satellite)](https://access.redhat.com/articles/3664871).
+
+## Commonly modified variables
+
+| Name                  | Type | Default value           | Description                                     |
+|-----------------------|------|-------------------------|-------------------------------------------------|
+| leapp_upgrade_type    | String  | satellite | satellite, cdn or rhui |
+| leapp_preupg_opt | String | --no-rhsm when leapp_upgrade_type is connected | Upstream repository usage - whether to use RHSM |
+| leapp_metadata_url      | String |  | URL to the leapp metadata, usually over https |
```
swapdisk, line 23: Better to reference the detailed explanation rather than the same thing explained differently in two places. 

```suggestion
| leapp_metadata_url      | String |  | See Requirements section above. |
```

roles/analysis/README.md
```diff
@@ -1,27 +1,65 @@
...
+| leapp_enable_repos_args | String |  | --enablerepo (leapp_repos_enabled) or blank |
```
swapdisk, line 24: Remove this. Role users should not be messing with it. 

```suggestion
```

roles/analysis/README.md
```diff
@@ -1,27 +1,65 @@
...
-|-------------------------|-----------------------|-----------------------------------------------------|
-| leapp_upgrade_type      | "satellite"           | Set to "cdn" for hosts registered with Red Hat CDN and "rhui" for hosts using rhui repos. |
-| leapp_metadata_url      |                       | See Requirements section above.                     |
-| leapp_answerfile        |                       | Optional multi-line string. If defined, this will be used as the contents of `/var/log/leapp/answerfile`. |
-| leapp_preupg_opts       |                       | Optional string to define command line options to be passed to the `leapp` command when running the pre-upgrade. |
-| post_reboot_delay       | 120                   | Optional integer to pass to the reboot post_reboot_delay option. |
+## Variables for Satellite based upgrades
+| Name                  | Type | Default value           | Description                                     |
+|-----------------------|------|-------------------------|-------------------------------------------------|
+| satellite_organization  | String | Example | Organization used in Satellite definition |
```
swapdisk, line 30: "Example" is not the default value. That line in defaults/main.yml is commented out. 

```suggestion
| satellite_organization  | String |  | Organization used in Satellite definition |
```

roles/analysis/README.md
```diff
@@ -1,27 +1,65 @@
...
+| satellite_activation_key_leapp | String |  | Key used to identify activation key |
```
swapdisk, line 31: Change like this... 

```suggestion
| satellite_activation_key_leapp | String |  | Satellite activation key |
```

roles/analysis/README.md
```diff
@@ -1,27 +1,65 @@
...
+| satellite_activation_key_pre_leapp | String |  | initial state of subscriptions and svc level |
+| satellite_activation_key_leapp     | String |  | Post-IPU state of subscriptions and svc level |
```
swapdisk, line 33: I have no idea what this means. If role users really need to use these variables, we need to explain them much better than a short description that will fit in this table. Also, s/IPU/upgrade/g.

roles/analysis/README.md
```diff
@@ -1,27 +1,65 @@
...
+| leapp_repos_enabled    | List | [] | Satellite repo for the satellite client RPM install |
```
swapdisk, line 34: Same question as above. If this needs to be set in some specific way for Satellite users, it needs to be explained in more detail.

roles/analysis/README.md
```diff
@@ -1,27 +1,65 @@
...
+
+| Name                  | Type | Default value           | Description                                     |
+|-----------------------|------|-------------------------|-------------------------------------------------|
+| result_filename | String |  | Path to file for the output of leapp |
+| analysis_packages_el6 | List | redhat-upgrade-tool | RPMS that need to be installed for IPU to RHEL7 |
+| analysis_repos_el6    | List | rhel-6-server-extras-rpms rhel-6-server-optional-rpms | Repo to be enabled for IPU to RHEL7  |
+| analysis_packages_el7 | List | leapp-upgrade             | RPMS that need to be installed for IPU to RHEL7 |
+| analysis_repos_el7    | List | rhel-7-server-extras-rpms | Repo to be enabled for IPU to RHEL7 |
+| analysis_packages_el8 | List | leapp-upgrade | RPMS that need to be installed for IPU to RHEL8 |
+| analysis_repos_el8 | List | rhel-7-server-extras-rpms | Repo to be enabled for IPU to RHEL7             |
```
swapdisk, line 46: Remove these from the doc. Should not be used by role users... 

```suggestion
```

roles/analysis/README.md
```diff
@@ -1,27 +1,65 @@
...
+| leapp_answerfile | TXT file | /var/log/leapp/answerfile | Optional - Source for Alternate AnswerFile needed during leapp process while upgrading  |
```
swapdisk, line 47: No, this is wrong. The variable has no default value and it doesn't define the filename of the answerfile. It is used to define the contents of the answerfile. Please fix as follows:

```suggestion
| leapp_answerfile | Multi-line String |  | If defined, this writen to `/var/log/leapp/answerfile` before generating the pre-upgrade report. |
```

swapdisk (changes requested):
Please take a crack at the suggested changes. Also, if role users will need to set the variables for Satellite, those variables must be explained in more detail under a dedicated section.

swapdisk:
Heads up, @jwesterd-redhat. We've eliminated the `leapp_metadata_url` variable with #106, so do the same here to avoid merge conflict.

roles/analysis/README.md
jwesterd-redhat: Removed Metadata references

roles/analysis/README.md
jwesterd-redhat: removed the Post-activation key and added section for the initial activation key and the leapp activation key,

roles/analysis/README.md
jwesterd-redhat: removd.

roles/analysis/README.md
jwesterd-redhat: ok

swapdisk (approved):

PR #104, opened 2023-07-07
Updated README files

scott-vick:
- Updated the documentation README files for the `upgrade` and the `parse_leapp_report roles`.


roles/upgrade/README.md
```diff
@@ -18,6 +21,15 @@ The `upgrade` role is used to kick off the Leapp OS upgrade on the target host.
 
 See [`upgrade.yml`](../../playbooks/upgrade.yml).
 
+```ansible
+- name: Perform OS upgrade
+  ansible.builtin.import_role:
+    name: infra.leapp.upgrade
+```
```
swapdisk, line 28: Let just link to the example playbook rather than adding this and giving two different examples... 
```suggestion
```

swapdisk (changes requested):
Just the one suggestion and then should be GTM.

swapdisk (approved):

PR #110, opened 2023-07-31
unset release for cdn, too

swapdisk:
@djdanielsson, as discussed.


djdanielsson (approved):
I think this will work fine, I normally put ( ) around each statement between or but I don't think it is required

PR #116, opened 2023-08-28
Minor adjustments for RHUI environments.

heatmiser:
- utilize shell: yum-config-manager with RHUI vs package module for other modes
- Include RHUI for repo disables


djdanielsson (approved):
LGTM

PR #117, opened 2023-08-28
Fix for Pathing issue with analysis

scott-vick:
- Added default variable "os_path"
- Added extra parameter to the leapp analysis shell command to override or add to the default $PATH on the target node


djdanielsson (changes requested):
need to add to upgrade as well

scott-vick:
Updated all shell tasks to include the new os_path variable.

djdanielsson (approved):
I think this should work fine

Issue #124, opened 2023-09-14
selinux_mode variable fails to set a value if the target server has selinux disabled


scott-vick:
The selinux_mode default variable uses a fact generated during the pre-ripu analysis. However if the server has selinux set to disabled, the correct fact (config_mode) is not created and therefore causes a failure during the selinux mode configuration tasks, due to a missing variable.

The facts generated when selinux is active (enforcing or permissive):
```
"ansible_selinux": {
      "config_mode": "enforcing",
      "mode": "enforcing",
      "policyvers": 31,
      "status": "enabled",
      "type": "targeted"
},
```

The facts generated when selinux is disabled:
```
"ansible_selinux": {
      "status": "disabled"
},
```

scott-vick:
Created fix and submitted PR# [https://github.com/redhat-cop/infra.leapp/pull/125](#125 )

PR #125, opened 2023-09-14
Set selinux_mode variable to default to "disabled" when a server has selinux disabled

scott-vick:
Corrects error generated when the config_mode Ansible fact is missing.


djdanielsson (approved):
LGTM

swapdisk:
Solves issue #124.

PR #130, opened 2023-09-21
Configure async/poll values via default vars

heatmiser:
Setting async/poll values as default vars will allow override timer values based on the end user's particular needs.


swapdisk:
May I suggest:
- document the new vars in the role README files
- add changelogs/fragments/async-timer-vars.yml

heatmiser:
@swapdisk Sure thing. More in a bit...

Issue #137, opened 2023-10-03
what versions are supported? 


alexxa:
I can't find in ReadMe what environments are supported please? Clouds? AWS, MSFT, GCP, and on-premise? Evrth? And what RHEL versions 6to7, 7to8, 8to9? Does it work for clusters? vanilla RHEL only or for example SAP too? =) As a user, the supported environment/scenarios would be my first question when checking on it.

swapdisk:
Hi Irina, thank you for raising the issue. This is good feedback. 

We can certainly cover these questions in the README. For example, we should make it clear that the collection does work for RHEL upgrades 6to7, 7to8, and 8to9. 

As far as the other questions, I guess we are kind of depending on users understanding what is supported by reading the Red Hat docs covering those topics. After all, the collection doesn't actually "do" the upgrades, but rather, is a convenient wrapper to automate the underlying tools that do, i.e., RUT in the case of 6to7 and Leapp for 7to8 and later.

Let me take a crack at updating the README to better address these questions or at least point users at the appropriate docs for the ultimate source of truth. I'll tag you in the PR to review. Thanks!

djdanielsson:
I wonder if we should just link the access articles that explain which versions that way we are not constantly updating it to match

swapdisk:
> hey, any update please?

I opened PR #139. Let me know if this works for you?

PR #139, opened 2023-11-30
add supported versions info

swapdisk:
@alexxa, let me know if this works for #137?
- Add section explaining supported RHEL versions and link to solution article with details of supported upgrade paths including minor versions and RHEL for SAP HANA 
- Fixed a typo


djdanielsson (approved):
LGTM

PR #143, opened 2023-12-08
add more caveats

swapdisk:
@alexxa, following up from our chat earlier. Kindly let me know if this satisfies you for #137.


djdanielsson (approved):
LGTM

README.md
```diff
@@ -27,6 +27,12 @@ The collection supports RHEL in-place upgrades for the following RHEL versions:
 
 The collection may be used for the RHEL upgrade paths and minor versions supported by the indicated upgrade utilities (Leapp or RUT). Refer the to Red Hat knowledge solution article [Supported in-place upgrade paths for Red Hat Enterprise Linux](https://access.redhat.com/articles/4263361) for the latest support details.
 
+Third-party products and packages are not upgraded. It is expected that the playbook including the `upgrade` role will perform the additional tasks required to handle the upgrade of any installed third-party tools and agents, for example [Veritas Cluster](https://www.veritas.com/support/en_US/doc/infoscale_wp_upgradewithRedHat), [SAP HANA](https://access.redhat.com/solutions/5154031), etc. Likewise for packages installed from non-RHEL repositories such as [Red Hat Software Collections](https://access.redhat.com/support/policy/updates/rhscl), [EPEL](https://docs.fedoraproject.org/en-US/epel/), [RPM Fusion](https://rpmfusion.org/), etc.
```
alexxa, line 30: I'm uncertain about mismatch of two sentences "Third-party products and packages are not upgraded" and then the opposite -  " playbook will perform <..> to handle the upgrade of any installed third-party tool and agents". There are two different things 1) upgrade of OS, and 2) upgrade of any third-party tools and agents, e.g SAP software/workloads. Those are different procedures and different tooling. This repo/playbook is for OS upgrade only, and particularly RHEL packages which shipped for SAP will be upgrade. So to avoid any confusion, I would keep it only as "It is expected that the playbook <...> will perform the upgrade of RHEL operating system. The installed applications and utilities, along with the configurations and preferences, are incorporated into the new version....<...>" 
On the contrary, EPEL or Fusion packages are not upgraded, but in that sentence, to me, it sounds like they do.

README.md
```diff
@@ -27,6 +27,12 @@ The collection supports RHEL in-place upgrades for the following RHEL versions:
...
+
+Having said that, most application workloads will still function correctly after a RHEL in-place upgrade if simply left untouched, thanks to [RHEL Application Compatibility](https://access.redhat.com/articles/rhel8-abi-compatibility) support.
```
alexxa, line 32: most application workloads sounds mysterious and questionable.

swapdisk: Fair enough. Is this better?... 

Having said that, many application workloads will benefit from [RHEL Application Compatibility](https://access.redhat.com/articles/rhel8-abi-compatibility) support such that they will still function correctly after a RHEL in-place upgrade if simply left untouched. Of course, the only way to know for sure is to run a test upgrade and then assess if there is any unexpected impact to your app. Pro tip: Test in your lower environments before moving on to production.

README.md
alexxa: is it possible to drop any wording about (how to) upgrade third party tooling? If a user upgrade an OS, it doesn't necessary mean they need/want to upgrade the third party tooling. E.g. in a SAP case, as I wrote earlier, upgrade of RHEL and upgrade of SAP installed there, are different procedures, different tooling, independent from each other. This repo for upgrade of OS only. I don't see how it can be tweaked to upgrade 3rd party (as SAP applications).

README.md
alexxa: Yes, it's better. However, if the 3rd party SW is supported on the target version, does it really matter how that target version has been 'created', by the a fresh install or by the upgrade? I like a simple and clear line in the leapp doc: "The installed applications and utilities, along with the configurations and preferences, are incorporated into the new version. " from https://access.redhat.com/documentation/en-us/red_hat_enterprise_linux/8/html-single/upgrading_from_rhel_7_to_rhel_8/index#ref_key-migration-terminology_upgrading-from-rhel-7-to-rhel-8

README.md
swapdisk: > is it possible to drop any wording about (how to) upgrade third party tooling?...

I would rather not drop it, but I have reworded to make it less confusing and moved under "Not in scope" section so it's clear we are not talking about what the roles in this collection do.

README.md
swapdisk: Removed mysterious wording and moved under "Not in scope" section.

alexxa:
LGTM. Thank you!

Issue #108, opened 2023-07-25
Update playbook example and other potential enhancements


ingmarfjolla:
Hi, we just used this collection to automate an upgrade and were wondering what enhancements might work / what might not be needed for the repo. @pavuluriharsha @elaczyns 

1. Updating the playbook examples for both the analysis and upgrade. 

I see that a PR with updated documentation has been merged, so would it be worthwhile to add example playbooks for the analysis and upgrade that make use of the variables? As an example in the upgrade playbook: 
```
- name: RHEL OS Upgrade
  hosts: all
  strategy: free
  force_handlers: true
  vars:
    satellite_organization: 'my_organization'
   satellite_activation_key_leapp: 'key'
   satellite_activation_key_post_leapp: 'key'
    leapp_upgrade_opts: '--target 8.8'
    update_grub_to_grub_2: true
    selinux_mode: permissive
 
  tasks:
    - name: Perform OS upgrade
      ansible.builtin.import_role:
        name: infra.leapp.upgrade
...
```
we needed to make use of the variables and having an example in the repo for others would be handy to see an example. 

2. Should this repo make mention of the xfs fix? 

One of the issues we ran into was here:  https://access.redhat.com/solutions/5057391 and we tested and verified the upstream RPM worked without needing to add any extra space and this collection was able to default to using that. Is that something this repo should mention as a possible workaround?  If so, there could be an extra role added (we just added a task) that might take care of installing that on the target machines with the caveat that it's currently unsupported. 

3. Should the analysis task fail if there are inhibitors? 

We added our own assertion at the end of the analysis task that would fail in the case of any inhibitors so that we wouldn't move forward with the upgrade. This one is more of a question, but for my own understanding is there a reason the analysis task doesn't currently aim to fail if there are any inhibitors?

djdanielsson:
1) good idea
2) no, because it is a temporary thing. at most maybe create an issue with a solution
3) that is up to the users, in some cases people may not want it to fail but that is why we provide a fact for users to easily use just the way you did.

swapdisk:
@ingmarfjolla, does merged PR #109 solve number 1? Anything else needed to close this issue?

PR #144, opened 2023-12-10
DRAFT: Support base os instances on AWS, Azure, and GCP

Rezney:
Support base os PAYG instances on AWS, Azure, and GCP by adding:

```
leapp_rhui_map:
  rhui-aws: leapp-rhui-aws
  rhui-azure: leapp-rhui-azure
  rhui-gcp: leapp-rhui-google
```

Technically, we could use e.g. "ec2_metadata_facts" but we would have to add new dependencies.

### DIFF
--- a/roles/analysis/defaults/main.yml
+++ b/roles/analysis/defaults/main.yml
@@ -12,6 +12,9 @@ analysis_packages_el7:
   # - leapp-upgrade-el7toel8
   # Try this instead.
   - leapp-upgrade
+  # In case you are on a PAYG instance using RHUI, you have to install one of the
+  # "leapp-rhui" rpms providing the target system content, keys, and certificates
+  - "{{ leapp_rhui_map[leapp_upgrade_type] if leapp_upgrade_type.startswith('rhui') else '' }}"
 
 analysis_packages_el8:
   # TODO: Seems to cause package dependency problems with post upgrade cleanup.
@@ -20,8 +23,9 @@ analysis_packages_el8:
   - leapp-upgrade
   # TODO: This should be part of inhibitor remediation.
   # - vdo
-  # TODO: only w/ rhui for aws
-  # - leapp-rhui-aws
+  # In case you are on a PAYG instance using RHUI, you have to install one of the
+  # "leapp-rhui" rpms providing the target system content, keys, and certificates
+  - "{{ leapp_rhui_map.leapp_upgrade_type if leapp_upgrade_type.startswith('rhui') else '' }}"
 
 analysis_repos_el6:
   - rhel-6-server-extras-rpms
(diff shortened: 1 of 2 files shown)


Rezney:
@swapdisk @heatmiser Please let me know if the approach is ok.

Couple of questions:
- `analysis_repos_el7` from "playbooks/analysis.yml" is set to `copr:copr.fedorainfracloud.org:group_oamg:leapp` to enable the repo but there is nothing enabling the Fedora Copr. Do we want to introduce it or there are other plans?
- `leapp_preupg_opts` from "playbooks/analysis.yml" seems to be overriding `leapp_upgrade_opts` from the defaults. What about introducing a list and extending it?

swapdisk:
> Please let me know if the approach is ok...

Yes, I like this approach. 

> `analysis_repos_el7` from "playbooks/analysis.yml" is set to `copr:copr.fedorainfracloud.org:group_oamg:leapp` to enable the repo but there is nothing enabling the Fedora Copr. Do we want to introduce it or there are other plans?

Right, I don't know what's going on with that. This is just an example playbook, but it still should make sense. Anybody?

> * `leapp_preupg_opts` from "playbooks/analysis.yml" seems to be overriding `leapp_upgrade_opts` from the defaults. What about introducing a list and extending it?

Sounds good if we can do it in a forward compatible way, i.e., not breaking existing playbooks folks may have that are already overriding `leapp_upgrade_opts`?

djdanielsson:
I believe the copr fedora stuff was an example but I am not positive @jeffmcutter do you remember?

jeffmcutter:
That was added after my time:

https://github.com/redhat-cop/infra.leapp/commit/7a647c8e2b13a9db8288a50e5d399e4b97bb4c4b

@ingmarfjolla Can you help shed some light on the question?

ingmarfjolla:
If I remember correctly at the time I was adding those examples, we were pointing to `copr:copr.fedorainfracloud.org:group_oamg:leapp` for 2 reasons. 1 was to just show the usage of that variable in general, and 2 we were testing an upstream fix for an issue we had with leapp at the time (I don't work at RH anymore but in the leapp gchat there should still be a a thread about it). Heres the original [issue](https://github.com/redhat-cop/infra.leapp/issues/108) where I made the suggestion

Rezney:
Thanks all for chiming in. I wanted to make sure I am not stepping on anyone's toes with enabling it. I believe it is really valuable for testing the upstream bits as @ingmarfjolla mentioned.

djdanielsson:
Oh I remember it was because the bug fix for xfs wasn't downstream yet

heatmiser:
@Rezney it is my opinion that we move away from requiring the end user to manually set `leapp_upgrade_type` and then rely on logic that stems from that setting. Instead, I feel that we should be programmatically introspecting each system and then setting the requisite repos that need to be enabled. I began this type of effort in the leapp-project we use for workshops, [here](https://github.com/redhat-partner-tech/leapp-project/blob/1f47474dc5679938386e00176cecb5884bf1a687/analysis.yml#L18-L45). We should be able to determine where/how a particular node/instance is running, review either current subscription mode or currently enabled repos to then set the appropriate repos needed for leapp execution. In this manner, we could have 5 different RHEL nodes in a particular inventory for an ansible leapp playbook, with some running on bare-metal, some on VMware, some on various clouds, subscribed viu Satellite or RH CDN or via RHUI...each node/instance could be handled correctly.

Rezney:
@heatmiser Nice! I was not aware we have "Amazon EC2" info for free and I counted with the natural reserve against bringing in more dependencies to get e.g. "ec2_metadata_facts". Do you happen to know we have similar info for Azure and GCP (lazy ask - I can check) ?
I definitely agree with the diverse inventory reasoning. Ok, let's go this way.

djdanielsson:
how is this coming along or are you still waiting for some answers?

Rezney:
I plan to resume working on this next week. We have been snowed under because of a new release. Sorry for the delay.

heatmiser:
> how is this coming along or are you still waiting for some answers?

I have been working on this intermittently over the past month. I am planning to submit a PR over the weekend.

Rezney:
> > how is this coming along or are you still waiting for some answers?
> 
> I have been working on this intermittently over the past month. I am planning to submit a PR over the weekend.

Sounds good to me. Closing this PR then.

PR #146, opened 2024-01-08
Fix Post-Upgrade Kernel Validation on Ansible 2.14+

blainestory:
When I upgrade a host, it fails during post-upgrade tasks with the following error:

    TASK [infra.leapp.upgrade : Validate running kernel matches OS version] ********
    fatal: [host01.example.org]: FAILED! => {"msg": "The conditional check ''el8' in ansible_kernel' failed. The error was: Conditional is marked as unsafe, and cannot be evaluated."}

I believe the issue stems from a vulnerability fix introduced in Ansible 2.14.12 (first bullet under Playbook section):
https://docs.ansible.com/ansible/latest/porting_guides/porting_guide_core_2.14.html#playbook

Wasn't sure if a new task was warranted as the workaround was to do another set_fact including the 'el' string.

### DIFF
--- a/roles/upgrade/tasks/upgrade-validation.yml
+++ b/roles/upgrade/tasks/upgrade-validation.yml
@@ -5,7 +5,6 @@
 - name: Determine rhel_dest_major_version
   ansible.builtin.set_fact:
     rhel_dest_major_version: "{{ ((ansible_facts.ansible_local.pre_ripu.distribution_major_version | int) + 1) | string }}"
-    rhel_dest_kernel_check: "el{{ ((ansible_facts.ansible_local.pre_ripu.distribution_major_version | int) + 1) | string }}"
 
 - name: Validate current OS major version
   ansible.builtin.assert:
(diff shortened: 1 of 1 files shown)


djdanielsson (changes requested):
This would make the assert not function as intended

swapdisk:
May a suggest rather than introducing a new `rhel_dest_kernel_check` fact, just change the assertion conditional at line 17 to this?... 
```
        that: "'el' ~ rhel_dest_major_version in ansible_kernel"
```

blainestory:
Thank you for the suggestion, @swapdisk. That was more in line with what I was hoping to achieve. This was the first I had seen the ~ character in YAML. 

I can confirm it works on Ansible 2.15. I was able to upgrade a test RHEL7 box with no issues.

swapdisk:
@blainestory, LGTM. Thank you for the contribution.

@djdanielsson, does this work for you now?

djdanielsson (approved):
yes

PR #152, opened 2024-01-23
feat: Add custom repositories support

gothicx:


roles/analysis/tasks/analysis-leapp.yml
```diff
@@ -30,17 +30,26 @@
...
     content: "{{ leapp_answerfile }}"
     dest: /var/log/leapp/answerfile
     owner: root
     group: root
-    mode: '0644'
+    mode: "0644"
   when: leapp_answerfile is defined
 
+- name: Populate custom repositories file
+  ansible.builtin.copy:
```
djdanielsson, line 45: ansible.builtin.yum_repository should be used here I think

gothicx: Yes, we can probably use that to create the repositories, but I was thinking more to use that in the upgrade phase and not in the analysis phase. I think the file leapp_upgrade_repositories.repo needs to be populated in plain text with all needed repositories in the analysis phase in order to have a SUCCESS.

I'll rework that and commit it to this PR.

jeffmcutter:
I think this might be addressed with https://github.com/redhat-cop/infra.leapp/pull/161.

djdanielsson:
I agree I think this ended up getting implemented in #161 so closing

PR #153, opened 2024-01-29
Allow leapp packages to be removed

gothicx:
Leapp packages that were installed with the Leapp utility are automatically added to the exclude list to prevent critical files from being removed 

https://access.redhat.com/documentation/en-us/red_hat_enterprise_linux/8/html/upgrading_from_rhel_7_to_rhel_8/performing-post-upgrade-tasks-rhel-7-to-rhel-8_upgrading-from-rhel-7-to-rhel-8#doc-wrapper

### DIFF
--- a/roles/upgrade/tasks/leapp-post-upgrade.yml
+++ b/roles/upgrade/tasks/leapp-post-upgrade.yml
@@ -41,6 +41,9 @@
 - name: Include check-for-old-packages.yml
   ansible.builtin.include_tasks: check-for-old-packages.yml
 
+- name: Delete exclude line from dnf.conf
+  ansible.builtin.command: "/bin/dnf config-manager --save --setopt exclude=''"
+
 - name: Remove leapp related packages
   ansible.builtin.package:
     name:


djdanielsson (commented):
I think changing from package to yum and then using excludes option if passing blank excludes then allows it to be removed is a lot better then using a command to do this. if that doesn't work (someone would need to test to verify) then there are some other changes that would be required for lint... speaking of which.... just realized that isn't working....

djdanielsson (changes requested):
see comment

djdanielsson:
yum module actually has disable_excludes option which I think is better, if you update to use that then I will approve

jeffmcutter:
This is already present.  https://github.com/redhat-cop/infra.leapp/blob/main/roles/upgrade/tasks/leapp-post-upgrade.yml#L3

swapdisk:
As Jeffery noted, the `Clean dnf exclude list` task earlier in this task file is already doing this.

PR #155, opened 2024-01-29
updating ansible-lint

### LINKED ISSUE #154: fix Ansible lint
I just realized, I think I messed up linting when I was trying to fix the release code... I will try and fix this ASAP

djdanielsson:
This is to fix #154 while I found it actually is probably working, it was out of date and I removed the pre-commit in favor of the Ansible-lint GitHub action that now seems to work better.


roles/upgrade/tasks/leapp-post-upgrade.yml
```diff
@@ -76,25 +80,26 @@
...
   ansible.builtin.include_tasks: update-and-reboot.yml
 
 # TODO: Validate RHEL OS versions again?
 
 # Only found in RHEL 7 to 8 documentation
-- name:  Old kernels have been removed from the bootloader entry for RHEL 7 to 8
+- name: leapp-post-upgrade |  Old kernels have been removed from the bootloader entry for RHEL 7 to 8
   ansible.builtin.shell: |
-    export PATH={{ os_path }}
+    set -o pipefail;
```
djdanielsson, line 91: this change here would be the only concern that probably should be tested to ensure that the export path doesn't need to be first and we can set pipefail before it without issue. I do not think it should matter but in theory that is the main change that I could see breaking something unexpected

jeffmcutter (approved):
Code LGTM.  I used `gh pr checkout 155` to test from 6-7, 7-8, and 8-9.  All successful.

Issue #134, opened 2023-09-27
Post upgrade does not provide a method to "lock" a node to a version of RHEL


scott-vick:
When running the post-upgrade for RHEL 7 -> 8, the leapp-post-upgrade role does not have a method to lock the version to a specific release. If I want to upgrade from RHEL 7 -> 8.6 the post-upgrade process upgrades to RHEL 8.8 which is undesirable. 

The leapp upgrade allows for a target version for the upgrade to be specified (e.g. "--target 8.6") but it does not allow for the target to remain at the specified release.

swapdisk:
Hi Scott, this seems like a reasonable feature to add. Are you planning to work on this and bring us a PR? Thanks in advance! :slightly_smiling_face:

heatmiser:
@scott-vick This functionality would need to operate slightly differently depending on which package manager repository method is being utilized. Essentially to utilize the specific target feature (e.g. "--target 8.6"), you are going to need to have a means for the target node to have repositories with the desired target version enabled (e.g. "8.6").

Have you tried pinning the release version for [RHSM](https://access.redhat.com/solutions/238533) or [RHUI](https://access.redhat.com/solutions/4251981)?

We can implement version pinning via Satellite...referring to these example variables for the upgrade role:
`satellite_activation_key_pre_leapp: RHEL7_Dev`
`satellite_activation_key_leapp: RHEL7IPU_Dev`
`satellite_activation_key_post_leapp: RHEL8_Dev`
...the `satellite_activation_key_leapp` activation key will be associated with a content view that contains the current RHEL release version repositories (e.g. "7.9"), as well as the desired upgrade target RHEL version repositories (e.g. "8.6"). And then to keep a system pinned to that target RHEL version, the `satellite_activation_key_post_leapp` variable would be set to an activation key/content view that has repositories defined for the final target version, in this case we would define the same "8.6" repositories.

Denney-tech:
I also would like to stop the leapp.upgrade role from unsetting the target release and then following up with another update to latest release. I'm trying to perform a leapp from 7.9 to 8.8 to 9.2, but it 7.9 to 8.8 to 8.9 to 9.2 is invalid because 8.9's only valid leapp target is 9.3 (as of this writing).  As is stands, I can leapp to 8.8 just fine, but then unexpectedly end up on 8.9 afterwards.

@heatmiser pinning the target version with RHSM doesn't work for a couple of reasons:
[...]

PR #159, opened 2024-03-12
Do not unset release if release target is defined

Denney-tech:
This is a proposed fix for #134.

### DIFF
--- a/roles/analysis/defaults/main.yml
+++ b/roles/analysis/defaults/main.yml
@@ -41,6 +41,8 @@ leapp_upgrade_type: satellite
 # TODO: Add support for custom repositories?
 # leapp_upgrade_type: custom
 
+leapp_target_release: ""
+
 leapp_preupg_opts: "{{ '--no-rhsm' if leapp_upgrade_type == 'rhui' else '' }}"
 
 # Satellite Organization and Activation Keys are required if using Satellite to change content views
--- a/roles/analysis/tasks/analysis-leapp.yml
+++ b/roles/analysis/tasks/analysis-leapp.yml
@@ -46,6 +46,7 @@
     set -o pipefail;
     export PATH={{ os_path }};
     leapp preupgrade
+    {{ "--target " ~ leapp_target_release if leapp_target_release | length > 0 }}
     {{ leapp_preupg_opts }}
     {{ leapp_enable_repos_args }}
     2>&1 | tee -a {{ log_file }}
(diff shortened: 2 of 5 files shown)


heatmiser:
Reviewers, please refer to [the issue here](https://github.com/redhat-cop/infra.leapp/issues/134) for additional background and discussion.

djdanielsson:
I think we would need to test the code but from just looking at it, I think it looks good

Denney-tech:
When I get a moment later today, I will add an assertion to make sure that `--target` isn't in the *_opts vars if leapp_target_release is also set.

That or, if preferred, we can just add a boolean variable that defaults to true for the unset task. Or we could even do both.

jeffmcutter:
I think it makes sense to include a task to set it to the desired destination release for the version of RHEL upgraded to as well.

Denney-tech:
@jeffmcutter do you mean having a separate `subscription-manager release --set x.y` task? Leapp itself sets the release to whatever it upgrades to, but I suppose there could also be a subsequent minor release target set after the major upgrade and before the minor upgrade steps.

jeffmcutter:
That is what I was thinking.  I'm not sure how common it would be, but with the right conditionals it would be harmless and potentially helpful.

Denney-tech:
@djdanielsson @heatmiser How's this?

Denney-tech:
Oh hold on, I forgot to add to README.md and add a changelog fragment.

jeffmcutter:
I believe this feature is going to be required for RHEL 8 High Availability cluster to RHEL 9.

Denney-tech:
Thank you for testing, @jeffmcutter. I just resolved some merge conflicts as well.

jeffmcutter:
> I believe this feature is going to be required for RHEL 8 High Availability cluster to RHEL 9.
> 
> I will test using this PR.

This is not the case specifically for the HA, I misread the docs on that.  Still testing.

jeffmcutter:
Tested successfully.  @swapdisk OK to merge?

Denney-tech:
Fixed sanity checks. I had created the change fragment in browser, which created it with dos line-endings instead of unix. Should be good now.

djdanielsson:
can you fix the conflict?

Denney-tech:
@djdanielsson done

djdanielsson (approved):
LGTM

PR #162, opened 2024-03-15
Allow skipping of validation of previous analysis results.

jeffmcutter:
This allows the approach of:
1. Run analysis and check if there are any unknown inhibitors not already known and handled by remediation.  If there are unknown inhibitors to the remediation automation, then those need to be addressed, if not...
2. Go straight to upgrade which includes snapshot, remediation, upgrade, etc. without running analysis again.

This is safe because the leapp upgrade will stop anyways if there are still inhibitors remaining after remediation.

This change does not change the existing default behavior, only allows an override of it.


djdanielsson (approved):
Like you pointed out I think this could reduce upgrade time by a decent amount if people are interested in that

PR #163, opened 2024-03-15
adding post_upgrade_update option

djdanielsson:
This should solve #134 or at least is needed to go with #159 if that ends up getting merged as well.


jeffmcutter:
LGTM and is functionality that would have been useful to me on my current project.

jeffmcutter (approved):
LGTM

PR #167, opened 2024-03-24
Fix analysis handler for registering to pre-leapp activation key.

jeffmcutter:
Was missing conditional and would error if satellite_activation_key_leapp is provided and not satellite_activation_key_pre_leapp.


djdanielsson (approved):
LGTM

Issue #165, opened 2024-03-20
Add option to unload pata_acpi prior to starting leapp upgrade.


jeffmcutter:
For a simpler way to work around this module getting reloaded despite blacklists etc.

I will work up a PR on this.  @scott-vick recently ran into this again, as did I.

jeffmcutter:
Solved by https://github.com/redhat-cop/infra.leapp/pull/168.

PR #168, opened 2024-03-25
Add option to unload kernel modules before running leapp upgrade.

jeffmcutter:
Add kernel_modules_to_unload_before_upgrade, an empty list by default.

kernel_modules_to_unload_before_upgrade: []

Can be useful for pesky kernel modules that get reloaded mysteriously, for example pata_acpi.


jeffmcutter:
Addresses https://github.com/redhat-cop/infra.leapp/issues/165.

djdanielsson:
Did you test against RHEL 6-7?

jeffmcutter:
I've been testing 6-7, 7-8, and 8-9 in parallel.

djdanielsson (approved):
LGTM

Issue #160, opened 2024-03-14
fata error on /var/log/ripu directory


gcccheng:
Hi,

We constantly receive below error message while running playbook to upgrade from rhel7 to rhel8

RUNNING HANDLER [infra.leapp.common : Add end time to log file] ****************
...
FAILED! => {"changed": false, "msg": "The destination directory (/var/log/ripu) is not writable by the current user. Error was: [Errno 13] Permission denied: b'/var/log/ripu/.ansible_tmpq3ycp1exripu.log'"}

I checked the role content, and found two relevant tasks

- name: Create new log file
  ansible.builtin.copy:
    content: |
      {{ job_name }}
      Job started at {{ now(fmt='%Y-%m-%dT%H:%M:%SZ', utc=true) }}
    dest: "{{ log_file }}"
    owner: root
    group: root
    mode: "0644"
  notify:
    - Archive log file

- name: Add end time to log file
  ansible.builtin.lineinfile:
    path: "{{ log_file }}"
    line: Job ended at {{ now(fmt='%Y-%m-%dT%H:%M:%SZ', utc=true) }}
    owner: root
    group: root
    mode: "0644"
  listen: Archive log file

The thing is that, Ansible job does not complain on the first task of creating the log file, however it complains on the second task which adds a line and  is a handler. 

We run the role with become: true option, so there should not be any permission issue.

So I wonder what could cause the error here? Thanks! @swapdisk @swapdisk @scott-vick

jeffmcutter:
Hi @gcccheng,

I have tested from 7-8 using root SSH login, passwordless user ssh and sudo, and with password for ssh and sudo.  The latter 2 of the 3 with root ssh disabled to ensure it's not using root.  I have been unable to replicate the issue.

Are you able to provide some additional details?  Perhaps the following?

df -h /var/log/ripu
ls -laZ /var/log/ripu

If using ansible-navigator or Ansible Controller, the json details for the task.  If not, run ansible-playbook with -vvvv and provide the output for that task.

Thanks,
-Jeff

gcccheng:
Hi @jeffmcutter , thanks for replying, check below  fresh error with -vvv and command result 

RUNNING HANDLER [infra.leapp.common : Add end time to log file] ****************
[...]

PR #175, opened 2024-04-15
Switch ansible_python_interpreter for 7 - 8 upgrades.

jeffmcutter:
Switch from /usr/bin/python3 to /usr/libexec/platform-python.

May address https://github.com/redhat-cop/infra.leapp/issues/160 and doesn't change behavior.

### DIFF
--- a/roles/upgrade/tasks/leapp-upgrade.yml
+++ b/roles/upgrade/tasks/leapp-upgrade.yml
@@ -99,9 +99,9 @@
   timeout: 43260
 
 # Ansible discovers /usr/bin/python on RHEL 7, then on RHEL 8 it is no longer available.
-- name: leapp-upgrade | Set python interpreter to /usr/bin/python3 since /usr/bin/python is no longer available
+- name: leapp-upgrade | Set python interpreter to /usr/libexec/platform-python since /usr/bin/python is no longer available
   ansible.builtin.set_fact:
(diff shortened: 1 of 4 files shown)


roles/upgrade/tasks/leapp-upgrade.yml
```diff
@@ -99,9 +99,9 @@
   timeout: 43260
 
 # Ansible discovers /usr/bin/python on RHEL 7, then on RHEL 8 it is no longer available.
-- name: leapp-upgrade | Set python interpreter to /usr/bin/python3 since /usr/bin/python is no longer available
+- name: leapp-upgrade | Set python interpreter to /usr/libexec/platform-python since /usr/bin/python is no longer available
   ansible.builtin.set_fact:
-    ansible_python_interpreter: /usr/bin/python3
+    ansible_python_interpreter: /usr/libexec/platform-python
```
djdanielsson, line 104: I am wondering if we should make this a variable that defaults to this so someone could override if they needed

djdanielsson (changes requested):
what do you think about this idea?

roles/upgrade/tasks/leapp-upgrade.yml
jeffmcutter: It doesn't hurt, that's for sure.

djdanielsson (approved):
LGTM

Issue #172, opened 2024-04-03
playbook hanging with reboot phase


gcccheng:
We are upgrading rhel7 to rhel8, and one of machines is hanging on the rebooting phase for 3 hours(normally 40 minutes), see below output from playbook job

..
TASK [infra.leapp.upgrade : Start Leapp OS upgrade] **************************** 
ASYNC POLL on lxrp001.example.com: jid=j641575276660.2466 started=1 finished=0
Waiting for job to be finished. Sleeping for 10 minutes... 
ASYNC POLL on  lxrp001.example.com: jid=j641575276660.2466 started=1 finished=0
..
ASYNC OK on  lxrp001.example.com: jid=j641575276660.2466 
changed: [ lxrp001.example.com]

TASK [infra.leapp.upgrade : Reboot to continue Leapp OS upgrade] ***************   
..
on the console of the machine, we could see a login interface with 
Red Hat Enterprise Linux 8.6 (Ootpa)
Kernel 4.18.0-372.32.1.2l8_6.x86_64 on an x86_64

It seems like the upgrading it partially finished and got stuck in the middle.

I checked the infra/leapp/roles/upgrade/tasks/leapp-upgrade.ym file, and it has timeout set as 43260, which is 12 hours.
- name: Start Leapp OS upgrade
  ansible.builtin.shell: >
    export PATH={{ os_path }};
    set -o pipefail;
    leapp upgrade
    {{ leapp_upgrade_opts }}
    {{ leapp_enable_repos_args }}
    2>&1 | tee -a {{ log_file }}
  args:
    executable: /bin/bash
  async: "{{ async_timeout_maximum | int }}"
  poll: "{{ async_poll_interval | int }}"

- name: Reboot to continue Leapp OS upgrade
  ansible.builtin.reboot:
    msg: "Host is starting Leapp OS upgrade now!"
    reboot_timeout: 43200
    post_reboot_delay: "{{ post_reboot_delay }}"
  timeout: 43260

 Is there any reason why the timeout is so long? Is there any case where upgrading did succeed with 12 hours?@jeffmcutter 
Thanks!

djdanielsson:
I think the original idea was to just set it to an extremely long time where if there is any chance that the system is functioning and might come back it would have the time to do so. we probably should variablize those timeouts to allow the users to set it to what they want to allow.

jeffmcutter:
[...]

PR #176, opened 2024-04-15
Add reboot and upgrade timeout variables

jeffmcutter:
Add task timeout to reboots without it.

Remove post_reboot_delay from analysis roll where there are no reboots.

Addresses https://github.com/redhat-cop/infra.leapp/issues/172.

### DIFF
--- a/changelogs/fragments/post_upgrade_update copy.yml
+++ b/changelogs/fragments/post_upgrade_update copy.yml
@@ -1,0 +1,4 @@
+---
+major_changes:
+  - Variabilize reboot_timeout and upgrade_timeout.
+...
--- a/roles/upgrade/tasks/redhat-upgrade-tool-upgrade.yml
+++ b/roles/upgrade/tasks/redhat-upgrade-tool-upgrade.yml
@@ -94,6 +94,8 @@
 - name: redhat-upgrade-tool-upgrade | Reboot to continue OS upgrade
   ansible.builtin.reboot:
     msg: Host is starting OS upgrade now!
+    reboot_timeout: "{{ upgrade_timeout }}"
     post_reboot_delay: "{{ post_reboot_delay }}"
-    reboot_timeout: 43200
+  timeout: "{{ upgrade_timeout }}"
+
 ...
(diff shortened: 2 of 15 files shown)


changelogs/fragments/post_upgrade_update copy.yml
```diff
@@ -0,0 +1,4 @@
+---
+major_changes:
```
djdanielsson, line 2: not sure this is really a major change, I think this is more a minor one.

djdanielsson (changes requested):
lets just change it to minor change in the logs and then it looks good

djdanielsson (approved):
LGTM

roles/upgrade/tasks/redhat-upgrade-tool-upgrade.yml
```diff
@@ -94,6 +94,8 @@
 - name: redhat-upgrade-tool-upgrade | Reboot to continue OS upgrade
   ansible.builtin.reboot:
     msg: Host is starting OS upgrade now!
+    reboot_timeout: "{{ upgrade_timeout }}"
     post_reboot_delay: "{{ post_reboot_delay }}"
-    reboot_timeout: 43200
+  timeout: "{{ upgrade_timeout }}"
```
joeashcraft, line 99: task keyword 'timeout' may not be present in Ansible 2.9, so this change takes infra.leapp to ansible-core 2.10+

evidence:  
see timeout task keyword in [stable-2.10](https://github.com/ansible/ansible/blob/05b90ab69a3b023aa44b812c636bb2c48e30108e/docs/docsite/keyword_desc.yml#L73) but not [stable-2.9](https://github.com/ansible/ansible/blob/5cfdbd83151edd000e79edf550f0c5bc54aebd19/docs/docsite/keyword_desc.yml#L70-L71)

djdanielsson: We do not support 2.9

jeffmcutter: We should update https://github.com/redhat-cop/infra.leapp/blob/main/meta/runtime.yml.

jeffmcutter: Do we still need this?  https://github.com/jeffmcutter/ansible-leapp/blob/main/.github/workflows/ansible-test.yml#L78

PR #178, opened 2024-04-16
Fix leapp_inhibitors for PR 177 when none found.

jeffmcutter:
Fix for https://github.com/redhat-cop/infra.leapp/pull/177.


djdanielsson (approved):
good catch

PR #179, opened 2024-04-17
Initial commit for HA upgrade playbook example.

jeffmcutter:
For RHEL HA 8.8+ - 9.2+ only as described in the link in the playbook.  It will not work for 7 to 8 as the underlying Leapp finds it as an inhibitor.

I'm looking for thoughts on where we might want to go with this.


djdanielsson (approved):
I think it is a good example of how you can do a cluster upgrade

alexxa:
@PRELVAS94 please see if it's any of your interest

Issue #182, opened 2024-04-25
enablerepo not a valid option for package module


k3mic:
roles/analysis/analysis-leapp.yml 

Seems like it might still work, but it's not mentioned at:

https://docs.ansible.com/ansible/latest/collections/ansible/builtin/package_module.html

Maybe consider ensuring the repo is enabled with `community.general.rhsm_repository` module before the package install task?

k3mic:
> or we can just use the yum/dnf modules which do support those options and why this hasn't been an issue before because package ends up calling those modules anyways.

Only concern is if the repo is enabled for a one time install of `leapp-upgrade` and it's deps. The newly installed package(s) may go unpatched in the future. Unless the packages are cleaned up after upgrade (haven't dug that deep yet, on mobile)...

jeffmcutter:
ansible.builtin.package as per the documentation "acts as a proxy to the underlying package manager module. While **all arguments will be passed to the underlying module**, not all modules support the same arguments. This documentation only covers the minimum intersection of module arguments that all packaging modules support."

This means you need to refer to the underlying module to know what options are valid.  They are not documented in the package module.

https://docs.ansible.com/ansible/latest/collections/ansible/builtin/package_module.html#synopsis

I have changed my activation key for RHEL 7 leapp upgrades to only enable the 1 main repo:

```
[root@satrhel7-1 ~]# subscription-manager repos --list-enabled
+----------------------------------------------------------+
    Available Repositories in /etc/yum.repos.d/redhat.repo
+----------------------------------------------------------+
Repo ID:   rhel-7-server-rpms
Repo Name: Red Hat Enterprise Linux 7 Server (RPMs)
Repo URL:  https://satcap1.localdomain.local/pulp/content/my_org/Dev/rhel7/content/dist/rhel/server/7/7Server/$basearch/os
Enabled:   1
[root@satrhel7-1 ~]# 
```

I uninstalled all leapp packages, and re-ran analysis using the EE from AAP 2.4, 2.3, and 2.1, uninstalling leapp after each time.  All 3 test OK.  The leapp packages get installed.  I verified during that only the one repo was still enabled.

```
TASK [infra.leapp.analysis : analysis-leapp | Install packages for preupgrade analysis on RHEL 7] ***
changed: [satrhel7-1.localdomain.local]
Friday 26 April 2024  01:42:10 +0000 (0:00:05.680)       0:00:23.296 **********
```

[...]

PR #183, opened 2024-04-25
fixing #182

djdanielsson:
this is a fix for the issue #182 there is no value in using package because Leapp only works on RHEL based systems so using yum and dnf will always work.

### DIFF
--- a/roles/analysis/tasks/analysis-leapp.yml
+++ b/roles/analysis/tasks/analysis-leapp.yml
@@ -18,14 +18,14 @@
   when: leapp_upgrade_type == "custom"
 
 - name: analysis-leapp | Install packages for preupgrade analysis on RHEL 7
-  ansible.builtin.package:
+  ansible.builtin.yum:
     name: "{{ analysis_packages_el7 }}"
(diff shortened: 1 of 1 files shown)


jeffmcutter:
Please see my reply on #182 .

https://github.com/redhat-cop/infra.leapp/issues/182#issuecomment-2078504666

In summary, unless we have evidence that package doesn't work, I think we should stay with package.  If we do have evidence that package doesn't work, then analysis-preupg.yml also has package with enablerepo and should also be changed.

djdanielsson:
I don't see a reason to keep package really, I think we could easily change them all to yum because that will use dnf if available. This isn't a multi OS automation it's strictly for RHEL

djdanielsson:
That being said, I would need to search where else we use package and change those if we do that so don't merge this

jeffmcutter:
Can you help me to understand what the problem is with package? Previously, I was instructed that it is the "best" option.  I've never seen a problem with it.  For #182 package wasn't the issue.

djdanielsson:
that is fine, I personally don't care that much either way.

PR #192, opened 2024-05-21
Allow providing a list of known inhibitors

### LINKED ISSUE #123: [RFE] Provide a method to list known findings
Leapp and infra.leapp preupgrade checks provide crucial findings for systems to be in-place upgraded. These findings must be investigated and a plan to address them in the most appropriate manner is needed. However, after a finding has been assessed and verified and the needed pre and/or post upgrade automation has been created it would be very helpful to allow listing the finding as "known/handled" and not report it for subsequent systems checked later on. This would allow fully concentrating on previously unseen findings or findings that require attention.

Consider a case where an organization has standardized to implement centralized logging with Splunk or endpoint monitoring with Tanium. In such a case it is expected that most if not all systems will have the splunkforwarder and TaniumClient RPMs installed thus when checking the readiness of thousands of systems for in-place upgrades receiving the same warning about these RPMs being installed does not provide much if any additional value. Or the finding about incompatibilities with grep(1) being reported, since grep is installed everywhere, having to observe this finding by infra.leapp for each and every system is not helpful.

The intra.leapp role collection should provide a method for a user to list known/handled findings so that once the user has addressed a finding it can then be listed as known/handled and won't be reported for other systems after that. This could provide notable time savings as when time goes by fewer and fewer findings will be reported as the list of known and handled findings grows.

Here's an example how this was done in a simple but very efficient playbook used for Leapp preupgrade checks:

https://github.com/myllynen/rhel-ipu-automation/blob/master/preupgrade.yml

The implementation in infra.leapp could be of course different but hopefully the above illustrates how such functionality could look like (see especially lines 50-171). Thanks.

myllynen:
Add a variable to allow providing a list of keys of known inhibitors, these known inhibitors will be ignored when setting leapp_inhibitors and considering the value for upgrade_inhibited.

Example usage, here VDO related inhibitors are listed as known and they will not cause the playbook to fail on any host:

```
    leapp_known_inhibitors:
      # vdo check
      - 2f9802dc91315806c7cdc3c18d7b74f2a2383285
      # vdo package
      - 429a99e13b19a7eebadbb8cb35233d8119bcf255
  tasks:
    - name: Generate preupgrade analysis report
      ansible.builtin.import_role:
        name: infra.leapp.analysis

    - name: Fail if unknown inhibitors found
      fail:
        msg: Unknown inhibitors found, see the logs for details.
      when: upgrade_inhibited
```

This is helpful when dealing with a large number of hosts and where remediations for the known inhibitors are available, for any new hosts only unknown or unhandled inhibitors would cause the playbook to fail and thus the need to investigate in detail.

Resolves #123

### DIFF
--- a/changelogs/fragments/allow_listing_known_inhibitors.yml
+++ b/changelogs/fragments/allow_listing_known_inhibitors.yml
@@ -1,0 +1,3 @@
+---
+minor_changes:
+  - Allow listing known inhibitors for which remediations are available
--- a/roles/analysis/defaults/main.yml
+++ b/roles/analysis/defaults/main.yml
@@ -52,6 +52,10 @@ leapp_repos_enabled: []
 # leapp_repos_enabled:
 #   - satellite-client-6-for-rhel-{{ ansible_distribution_major_version | int + 1 }}-x86_64-rpms
 
+# List of keys of known inhibitors which will be excluded when setting upgrade_inhibited and leapp_inhibitors
+# This allows ignoring inhibitors for which remediations are available and require no further attention
+leapp_known_inhibitors: []
+
 # For leapp_upgrade_type == "custom"
 # Used to configure repos before running leapp analysis / installing leapp packages.
 local_repos_pre_leapp: []
--- a/roles/analysis/README.md
+++ b/roles/analysis/README.md
@@ -39,9 +39,11 @@ See comments in defaults/main.yml for additional details.
 |-----------------------|------|-------------------------|-------------------------------------------------|
 | leapp_answerfile | Multi-line String |  | If defined, this is written to `/var/log/leapp/answerfile` before generating the pre-upgrade report. |
 | leapp_preupg_opts | String | | Optional string to define command line options to be passed to the `leapp` command when running the pre-upgrade. |
+| leapp_known_inhibitors | List | [] | List of keys of known inhibitors ignored when setting upgrade_inhibited and leapp_inhibitors. |
 | os_path | String | $PATH | Option string to override the $PATH variable used on the target node |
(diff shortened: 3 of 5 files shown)


roles/analysis/defaults/main.yml
```diff
@@ -52,6 +52,10 @@ leapp_repos_enabled: []
 # leapp_repos_enabled:
 #   - satellite-client-6-for-rhel-{{ ansible_distribution_major_version | int + 1 }}-x86_64-rpms
 
+# List of keys of known inhibitors which will be exluded when setting upgrade_inhibited and leapp_inhibitors
```
swapdisk, line 55: Fix typo... s/exluded/excluded/

swapdisk (changes requested):
Please fix typo and add a change description under `changelogs/fragments`.

roles/analysis/defaults/main.yml
myllynen: Thanks, fixed both.

changelogs/fragments/allow_listing_known_inhibitors.yml
```diff
@@ -0,0 +1,5 @@
+---
+minor_changes:
+  - Move collection dependencies from galaxy.yml to requirements.yml
```
swapdisk, line 3: Remove 2 lines above. Cut and paste error? :scissors:

swapdisk (changes requested):

changelogs/fragments/allow_listing_known_inhibitors.yml
myllynen: D'oh.. Fixed now!

PR #195, opened 2024-05-22
Use ansible.builtin.dnf

### LINKED ISSUE #194: CI workflow broken by upstream release of ansible-core 2.17
With the 2024-05-20 release of ansible-core 2.17 (https://github.com/ansible/ansible/pull/83280), we find this under the **Removed Features** section of [CHANGELOG-v2.17.rst](https://github.com/ansible/ansible/blob/stable-2.17/changelogs/CHANGELOG-v2.17.rst): 

- Removed Python 2.7 and Python 3.6 as a supported remote version. Python 3.7+ is now required for target execution.
- With the removal of Python 2 support, the yum module and yum action plugin are removed and redirected to `dnf`.

I believe it is this change that is triggering the ansible-lint errors were are now seeing with our CI workflow. For example, from CI job [here](https://github.com/redhat-cop/infra.leapp/actions/runs/9179237705/job/25240931430#step:7:218):
```
fqcn[action-core]: Use FQCN for builtin module actions (ansible.builtin.yum).
roles/remediate/tasks/leapp_missing_efibootmgr.yml:4 Use `ansible.builtin.dnf` or `ansible.legacy.dnf` instead.

fqcn[action-core]: Use FQCN for builtin module actions (ansible.builtin.yum).
roles/remediate/tasks/leapp_multiple_kernels.yml:29 Use `ansible.builtin.dnf` or `ansible.legacy.dnf` instead.

fqcn[action-core]: Use FQCN for builtin module actions (ansible.builtin.yum).
roles/remediate/tasks/leapp_unavailable_kde.yml:4 Use `ansible.builtin.dnf` or `ansible.legacy.dnf` instead.

Error: Use FQCN for builtin module actions (ansible.builtin.yum).
Error: Use FQCN for builtin module actions (ansible.builtin.yum).
Error: Use FQCN for builtin module actions (ansible.builtin.yum).
Read documentation for instructions on how to ignore specific rule violations.

                 Rule Violation Summary                  
 count tag               profile    rule associated tags 
     3 fqcn[action-core] production formatting           

Failed: 3 failure(s), 0 warning(s) on 122 files. Profile 'production' was required, but 'shared' profile passed. Rating: 4/5 star
```
I'm not sure how we should deal with this. We would prefer to keep the CI workflows compatible with the latest upstream Ansible releases, but we also need the collection to be compatible back to older OS versions where we don't have Python 3 and `dnf`. If we simply change `ansible.builtin.yum` to `ansible.builtin.dnf`, does that still work on EL6? 

I'm open to suggestions. @djdanielsson or @jeffmcutter?

swapdisk:
Resolves #194.

### DIFF
--- a/roles/remediate/tasks/leapp_missing_efibootmgr.yml
+++ b/roles/remediate/tasks/leapp_missing_efibootmgr.yml
@@ -2,7 +2,7 @@
 - name: leapp_missing_efibootmgr | Install efibootmgr package - required on EFI systems
   block:
     - name: leapp_missing_efibootmgr | Install efibootmgr package
-      ansible.builtin.yum:
+      ansible.builtin.dnf:
         name: efibootmgr
         state: present
 
--- a/roles/remediate/tasks/leapp_multiple_kernels.yml
+++ b/roles/remediate/tasks/leapp_multiple_kernels.yml
@@ -27,7 +27,7 @@
       when: installed_kernels.stdout_lines[0] != current_kernel.stdout
 
     - name: leapp_multiple_kernels | Remove old kernels
-      ansible.builtin.yum:
+      ansible.builtin.dnf:
         name: kernel-core-{{ item }}
         state: absent
       loop: "{{ installed_kernels.stdout_lines[1:] }}"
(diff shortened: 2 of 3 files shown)


djdanielsson (changes requested):
I think this will break on older systems, we should test to ensure this doesn't cause any unexpected issues

swapdisk:
> I think this will break on older systems, we should test to ensure this doesn't cause any unexpected issues

The changes are all in the `remediate` role which is only supported for RHEL8 and later, so it should be fine.

djdanielsson (approved):

PR #202, opened 2024-06-14
updated ansible versions in meta and updated ansible-test workflow

djdanielsson:
modified:   roles/analysis/meta/main.yml
modified:   roles/common/meta/main.yml
modified:   roles/parse_leapp_report/meta/main.yml
modified:   roles/remediate/meta/main.yml
modified:   roles/upgrade/meta/main.yml
modified:   .github/workflows/ansible-test.yml
modified:   roles/analysis/files/check-inodes.sh

PR #209, opened 2024-07-11
Fix check-inodes.sh script.

jeffmcutter:
Output from running analysis role:
```
line 36: badinitio: unbound variable
```

Output testing script manually:
```
[root@satrhel6-1 ~]# /tmp/check-inodes.sh 
/tmp/check-inodes.sh: line 36: badinitio: unbound variable
[root@satrhel6-1 ~]# bash /tmp/check-inodes.sh 
/tmp/check-inodes.sh: line 36: exit: : numeric argument required
```

### DIFF
--- a/changelogs/fragments/fix_check_inodes.yml
+++ b/changelogs/fragments/fix_check_inodes.yml
@@ -1,0 +1,2 @@
+minor_changes:
+  - Fix check-inodes.sh script introduced error.
--- a/roles/analysis/files/check-inodes.sh
+++ b/roles/analysis/files/check-inodes.sh
@@ -1,4 +1,6 @@
-#!/bin/bash
+#!/bin/bash -eu
+
+badinitio=0
 
 # Total count of local files
 total_inodes=$(df -P --local --inodes --total | tail -1 | awk '{print $3}')
@@ -33,4 +35,4 @@ for path in "${paths[@]}"; do
   fi
 done
 
-exit $badinitio
+exit "$badinitio"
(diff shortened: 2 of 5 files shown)


jeffmcutter:
@djdanielsson Can you look at this one?  https://github.com/redhat-cop/infra.leapp/pull/202 broke check-inodes.sh (see description above).  Thanks!

djdanielsson (approved):
Hum... I wondered if those changes would break it. I will have to figure out how to make it skip this in sanity check

djdanielsson (changes requested):
I'll try and fix this

swapdisk (commented):
The "unbound variable" error is because `$badinitio` is unset if the report shortfalls loop doesn't find any problems. It was fine until `-u` bash option was added. Per bash man page, that option tells bash to treat unset variables as an error...
```
-u   Treat unset variables and parameters other than the special para‐
     meters "@" and "*", or array variables subscripted  with  "@"  or
     "*",  as an error when performing parameter expansion.  If expan‐
     sion is attempted on an unset variable or  parameter,  the  shell
     prints  an  error  message, and, if not interactive, exits with a
     non-zero status.
```
Let me recommend we just add `badinitio=0` before the loop at line 23 and we should be able to leave everything else as it was and not need to do the sanity test ignores.

djdanielsson:
yep I figured that out and did just that

swapdisk (approved):

djdanielsson (approved):

jeffmcutter:
Pulled the latest main and ran analysis successfully on a RHEL 6 box.  :-)

PR #211, opened 2024-07-11
Move fact capture to common.

jeffmcutter:
In order to avoid potential drift of the data between analysis and upgrade for debugging issues during upgrade.


djdanielsson (approved):
If this was just for trying to run without analysis I would probably push back but I do think there is a case to be made that it should happen right before the upgrade as you said to prevent the chance of drift

Issue #215, opened 2024-08-08
Loop running leapp_known_inhibitors failing to due key not being included in data structure created for high (errors)


aharold92:
We are receiving this error after we started using the infra.leapp 1.3.1 version

`
FAILED! => {"msg": "The conditional check 'item.key not in leapp_known_inhibitors' failed. The error was: error while evaluating conditional (item.key not in leapp_known_inhibitors): 'dict object' has no attribute 'key'. 'dict object' has no attribute 'key'\n\nThe error appears to be in '/runner/requirements_collections/ansible_collections/infra/leapp/roles/analysis/tasks/analysis-leapp.yml': line 96, column 3, but may \nbe elsewhere in the file depending on the exact syntax problem. \n\n The offending line appears to be:\n\n-name: analysis-leapp | Capture inhibitors in a list leapp_inhibitors\n ^ here\n"} 
`

This error occurs because the data structure created by the task ` - name: Construct a data structure to add high errors to leapp_report_json` does not include a key in the data structure. 

That exclusion then leads to a data structure that looks like the following
```        {
          "audience": "sysadmin",
          "flags": [
            "inhibitor"
          ],
          "severity": "high",
          "summary": "Risk Factor: high (error)\nTitle: Not enough space available on /var/lib/leapp/scratch: Needed at least 4992 MiB.\nSummary: {\"detail\": \"The file system hosting the /var/lib/leapp/scratch directory does not contain enough free space to proceed all parts of the in-place upgrade. Note the calculated required free space is the minimum derived from upgrades of minimal systems and the actual needed free space could be higher.\\nNeeded at least: 4992 MiB.\\nSuggested free space: 5992 MiB (or more).\"}\nKey: 0f6a4d531ad6079bd7a85b02bc922ccc7c143312\n----------------------------------------",
          "title": "Likely issues found with high (error) severity found, see summary.",
          "hostname": "tst-ssplappi109.bbtnet.com"
        } 
```

This data structure is then added to `leapp_report_json`.

Then when the role reaches the following task
```
- name: analysis-leapp | Capture inhibitors in a list leapp_inhibitors
  ansible.builtin.set_fact:
    leapp_inhibitors: "{{ leapp_inhibitors | default([]) + [item] }}"
  when:
    - item.key not in leapp_known_inhibitors
    - (leapp_high_sev_as_inhibitors | bool and item.severity == 'high') or
      'inhibitor' in item.flags
[...]

Issue #170, opened 2024-03-25
Leapp reporting schema update for infra.leapp


jeffmcutter:
https://github.com/redhat-cop/infra.leapp/pull/169

Using reporting schema 1.2.0 `leapp --report-schema 1.2.0` may alleviate the need to parse the txt report for errors.  However it may have implications on other things that we should consider first.  Opening this issue to give this some visibility and so we can discuss.

swapdisk:
Fixed with #216.

PR #216, opened 2024-08-09
Use new Leapp report schema 1.2.0

jeffmcutter:
Addresses https://github.com/redhat-cop/infra.leapp/issues/170 and https://github.com/redhat-cop/infra.leapp/issues/215.

### DIFF
--- a/roles/parse_leapp_report/tasks/main.yml
+++ b/roles/parse_leapp_report/tasks/main.yml
@@ -25,8 +25,8 @@
   when:
     - item.key not in leapp_known_inhibitors | default([], true)
     - (leapp_high_sev_as_inhibitors | default(false, true) | bool and item.severity == 'high') or
-      'inhibitor' in item.groups or
-      'error' in item.groups
+      'inhibitor' in item.groups | default([], true) or
+      'error' in item.groups | default([], true)
   loop: "{{ leapp_report_json.entries }}"
 
 - name: Collect inhibitors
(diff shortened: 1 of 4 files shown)


jeffmcutter:
Found an issue.  Hang tight.

jeffmcutter:
False alarm.  I thought of another test to do, but I had a var wrong.  It's working.

roles/parse_leapp_report/tasks/main.yml
```diff
@@ -21,10 +21,12 @@
 - name: Check for inhibitors
   ansible.builtin.set_fact:
     upgrade_inhibited: true
+    leapp_inhibitors: "{{ leapp_inhibitors | default([]) + [item] }}"
   when:
     - item.key not in leapp_known_inhibitors | default([], true)
     - (leapp_high_sev_as_inhibitors | default(false, true) | bool and item.severity == 'high') or
-      'inhibitor' in item.flags
+      'inhibitor' in item.groups or
+      'error' in item.groups
```
swapdisk, line 29: Hit a `'dict object' has no attribute 'groups'` failure with this task testing on RHEL8 because one of the finding entries did not have a `groups` key. I asked my Leapp friends and was referred [here](https://github.com/oamg/schema-test/blob/d4d8a14cc998e9ba67f3f1ec8af55a2786798a6e/report-schema-v120.json#L181). The `groups` key is not required/mandatory, so when it's empty, it's not part of the entry. I think this would fix it... 
```suggestion
      'inhibitor' in item.groups | default([], true) or
      'error' in item.groups | default([], true)
```

swapdisk (changes requested):
Looks good so far except for this one fail with the "Check for inhibitors" task.

swapdisk (approved):
Tested generating reports and doing upgrades from 7.9, 8.8 and 8.10.

LGTM!

PR #219, opened 2024-08-22
increase per-process file descriptor limit

### LINKED ISSUE #218: Leapp fails with "unable to open database file" error
Leapp preupgrade and upgrade may fail with "unable to open database file" error when it exceeds per-process file descriptor limit as reported with issue [RHEL-26459](https://issues.redhat.com/browse/RHEL-26459). The recommendation documented [here](https://access.redhat.com/solutions/6878881) is to increase the limit using `ulimit -n 16384`.

Let's add this to the tasks that run `leapp` command in the analysis and upgrade roles.

swapdisk:
Fixes #218. 

Tested successfully in our workshop lab environment with reproducer known to trigger the issue.

### DIFF
--- a/changelogs/fragments/fix_ulimit.yml
+++ b/changelogs/fragments/fix_ulimit.yml
@@ -1,0 +1,4 @@
+---
+bugfixes:
+  - Run leapp with increased per-process file descriptor limit
+...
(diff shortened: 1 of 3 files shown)


djdanielsson (approved):
LGTM

PR #224, opened 2024-09-30
Support optional gpgkey and repo_gpgcheck yum repository attributes

rjo-uk:
This PR adds support for the optional `gpgkey` and `repo_gpgcheck` yum repository attributes so that they can be included in the analysis and upgrade roles.  Support for the attributes has been added with the 'omit' default, so as not to impact existing users who are not using this feature.  The attributes are passed to the `ansible.builtin.yum_repository` module at the appropriate places.


swapdisk (approved):
LGTM.

PR #227, opened 2024-10-04
Add option to skip the update/reboot on pre-upgraded server

rjo-uk:
As part of the upgrade role, an initial update and reboot is performed in https://github.com/redhat-cop/infra.leapp/blob/main/roles/upgrade/tasks/leapp-upgrade.yml

Whilst this is good practice, it would be useful to have the option of disabling this so that:

- Any updates that are available on the server are not automatically applied.  For example, updates from a third party repository may bring in unwanted packages (eg a new third party kernel) and result in a server that cannot be upgraded
- Later code in the playbook may fail gracefully (for example, due lack of disk space) before the upgrade is performed.  If no additional updates have been installed on the current operating system, it is easier to recover the server to the pre-upgrade state.
- The pre-upgrade update can be performed during a different change window ahead of upgrade work.  This will reduce upgrade time by not having to perform a reboot.

This initial draft is intended to not change workflows for existing users.  If the variable `skip_pre_upgrade_update_reboot` is not provided, the update and reboot will occur as it does today.

### DIFF
--- a/roles/upgrade/tasks/leapp-upgrade.yml
+++ b/roles/upgrade/tasks/leapp-upgrade.yml
@@ -29,7 +29,7 @@
 
 - name: leapp-upgrade | Include update-and-reboot.yml
   ansible.builtin.include_tasks: update-and-reboot.yml
-  when: skip_pre_upgrade_update_reboot is not defined or not skip_pre_upgrade_update_reboot|bool
+  when: pre_upgrade_update | bool
 
 - name: leapp-upgrade | Create /etc/leapp/files/leapp_upgrade_repositories.repo
   ansible.builtin.yum_repository:
--- a/roles/upgrade/tasks/leapp-upgrade.yml
+++ b/roles/upgrade/tasks/leapp-upgrade.yml
@@ -29,6 +29,7 @@
 
 - name: leapp-upgrade | Include update-and-reboot.yml
   ansible.builtin.include_tasks: update-and-reboot.yml
+  when: skip_pre_upgrade_update_reboot is not defined or not skip_pre_upgrade_update_reboot|bool
 
 - name: leapp-upgrade | Create /etc/leapp/files/leapp_upgrade_repositories.repo
   ansible.builtin.yum_repository:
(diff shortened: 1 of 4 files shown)


roles/upgrade/tasks/leapp-upgrade.yml
```diff
@@ -29,6 +29,7 @@
 
 - name: leapp-upgrade | Include update-and-reboot.yml
   ansible.builtin.include_tasks: update-and-reboot.yml
+  when: skip_pre_upgrade_update_reboot is not defined or not skip_pre_upgrade_update_reboot|bool
```
jeffmcutter, line 32: When would the var not be defined?

jeffmcutter: Seems reasonable to me otherwise.  I would want to remove the check for defined since it's in the defaults.

rjo-uk: Hi @jeffmcutter 

Thanks for the feedback.  You're right that it's in the defaults so the check should not be needed.  Unfortunately, in the past I've known users to update individual task files from git rather than pulling all the updates (yes, that's bad and not logical...but that's how I've seen it done).  So someone just pulling the updated task file wouldn't have `skip_pre_upgrade_update_reboot` defined.  I agree that's bad practice, so I'll follow your advice and remove that check.

Before I do, there is another option:

Rather than defining `skip_pre_upgrade_update_reboot` I could write the logic the other way around and instead have something like`pre_upgrade_update` with a default of true.  Right now you have `post_upgrade_update` so maybe a `pre_upgrade_update` fits better?

Happy with either option as both achieves the same result.

jeffmcutter: Hi @rjo-uk,

That sounds good.  Consistent is always good, plus I think it will make the logic more straight forward.

Thanks,
-Jeff

swapdisk (approved):
I like the `pre_upgrade_update` var name.

LGTM

rjo-uk:
Changed from draft to final if you're happy with it.

PR #230, opened 2024-10-11
Updating the examples from 8.8 to 8.10

bontreger:
As RHEL 8.10 is the default and 8.8 is in EUS, updating the examples in the docs to point to 8.10


djdanielsson (approved):
LGTM

PR #231, opened 2024-11-13
Add result_fact_cacheable var

kalsto:
Allow AAP to cache facts from the leapp report in order to use it in future logic


swapdisk (approved):
LGTM

PR #234, opened 2024-11-15
leapp-report no longer uses groups - changed to flags

kalsto:
results_inhibitors was not being populated due to a change in the leapp report. It no longer uses "groups" as an identifier and uses "flags" instead.


swapdisk:
Hmm, I don't think so. Since the collection switched to using report schema 1.2.0 as was implemented with #216, you should be seeing "groups" instead of "flags" and "tags" in your leapp-report.json file.

kalsto:
Yep - my fault.. the report was generated without hte updated report-schema. Closing the PR.

PR #237, opened 2025-02-27
Add support for RHEL 7 remediate

### LINKED ISSUE #184: [RFE] remediate role support for RHEL 7
Hi!

any chance to use remediation on rhel7?

Thanks!

bontreger:
To resolve #184 

I do recommend that this increments the collection to 1.5.0 as it is a functionality addition


swapdisk (changes requested):
Thank you for the PR!

Please undo the changes to `CHANGELOG.rst` and `galaxy.yml` files as these get updated automatically by our release workflow. Instead, record your changes in a new yaml file under the `changelogs/fragments` directory. See previously merged PRs for examples.

In the new `leapp_pam_tally2.yml` file, fix the Ansible Lint "No new line character at the end of file" finding.

The rest of this looks reasonable to me. Give me a little time to play with it. I'll follow up to let you know how things go with my testing.

CHANGELOG.rst
```diff
@@ -3,7 +3,6 @@ Ansible Leapp Collection Release Notes
 ======================================
 
 .. contents:: Topics
-
```
swapdisk, line 6: Put back this blank line so `CHANGELOG.rst` remains untouched.

swapdisk:
As promised, I've started looking at this. Unfortunately, I didn't get very far with my testing before discovering a big problem with this role. It's not related to the changes you made, but rather the original implementation. 

What I found is that a number of the `leapp_*` playbooks use `ansible.builtin.meta: end_host` tasks when a condition is met indicating that remediation doesn't need to be run. This design was probably fine with Insights where I'm assuming each one was run as a distinct job. The problem is that now that these playbooks are being included en masse by the role, the first one to hit this condition results in the whole job ending right there rather than then job continuing to try the remaining playbooks listed in the `remediation_todo` var.

So it looks to me like we need to refactor the playbooks doing `ansible.builtin.meta: end_host` to instead use conditional blocks. Let me know if this makes sense or if you can think of a better way to do it?

roles/remediate/tasks/leapp_pam_tally2.yml
```diff
@@ -0,0 +1,20 @@
...
+        paths: /etc/pam.d/
+        file_type: file
+        contains: "pam_tally2.so"
+        read_whole_file: true
+        recurse: true
+      register: file_list
+
+    - name: leapp_pam_tally2 | Comment out all instances of pam_tally2 # noqa: command-instead-of-module
+      ansible.builtin.command:
+        cmd: "sed '/^#/! {/pam_tally2/ s/^/#/}' < {{ item }}"
```
swapdisk, line 17: Looks to me like this is just going to write the changes to stdout, but not make changes to the files. Maybe this instead?... 
```suggestion
        cmd: "sed -i '/^#/! {/pam_tally2/ s/^/#/}' {{ item }}"
```

bontreger: Thanks for the catch, was trying to write it in a way that would work with ansible.builtin.replace and wasn't getting it to cooperate with the two commands (so I don't add additional comments when it reruns)

Working on refactoring out the end_hosts now

bontreger:
Yeah, fortunately it's already in blocks, so I'm thinking I'll just set a fact (leapp_report_missing), use a failed_when and a rescue to kick us out of the current task block without stopping the whole run.  I'll have that ready shortly

swapdisk:
My apologies for the delay. I'm finally dug out from my other work and getting back to reviewing your latest changes today. Stay tuned!

swapdisk (approved):
@bontreger, thank you for your patience while we worked through dealing with all the problems uncovered that had nothing to do with your changes! :joy:

Just to summarize, we found that many of the playbooks in this role previously would prematurely end jobs, were not idempotent, did `cmd: sudo ...` which makes no sense since we run `become: true`, and on and on. So in addition to making the role now useful for 7 to 8 upgrades, we also cleaned up all that cruft.

I've finished testing in my lab on both RHEL7 and RHEL8 with `remediation_todo` listing all of the available remediate playbooks. Everything now appears idempotent, that is, after running on the same host twice, the second job summary always shows `changed=0`. I also tested running both before and after a Leapp preupgrade report was generated and no failed jobs either way.

So I'm calling this LGTM.

swapdisk:
Fixes #184.

swapdisk:
Fixes #240.

PR #238, opened 2025-03-11
Use community.general modules instead of modprobe command 

kubealex:
Use community.general.modprobe to remove unsupported modules in roles/remediate/tasks/leapp_loaded_removed_kernel_drivers.yml


swapdisk (approved):
LGTM. Thanks!

PR #241, opened 2025-03-23
Implement post-upgrade release set/unset for RHUI

### LINKED ISSUE #235: RHUI - Unset release during post-upgrade
For Leapp 8to9 for ARM architecture, 8.10 to 9.5 is currently inhibited. Instead, `--target 9.4` must be passed as a parameter, which is accomplished by setting `leapp_preupg_opts` and `leapp_upgrade_opts` for the leapp roles.

When using RHUI, this limitation leaves these hosts at 9.4, so I'd like to disucss a solution for unsetting the RHEL release version during post-upgrade for RHUI users so the hosts can be updated to 9.5 after the upgrade:
https://github.com/redhat-cop/infra.leapp/blob/main/roles/upgrade/tasks/leapp-post-upgrade.yml#L82

After the leapp upgrade, a release version is specified at `/etc/yum/vars/releasever`. We can easily clear this file using the ansible `file` module. In AWS, a tool is provided called `rhui-set-release` that helps set and unset this file, but it is just for convenience and we shouldn't need to use it in `infra.leapp.upgrade`.

We should also provide a post-upgrade set release task as well, to mirror to the subscription-manager feature. This would use the `file` module to set the yum variable to the value of `post_upgrade_release`.

This assumes that the "RHUI" repos that are defined are leveraging the `releasever` variable. Otherwise I suppose the user will need to specify custom repos.

Let me know if you have any feedback on this approach.

bparry02:
By default, RHUI repositories are no longer version-locked after an upgrade and will be updated to the latest available content. This behavior can be controlled using `post_upgrade_*` variables.

resolves #235

### DIFF
--- a/changelogs/fragments/unset-rhui-release.yml
+++ b/changelogs/fragments/unset-rhui-release.yml
@@ -1,0 +1,5 @@
+---
+major_changes:
+- Implement post-upgrade release version set/unset for 'RHUI' upgrade type.
+  By default, RHUI repositories are no longer version-locked after an upgrade
+  and will be updated to the latest available content.
--- a/roles/upgrade/tasks/leapp-post-upgrade.yml
+++ b/roles/upgrade/tasks/leapp-post-upgrade.yml
@@ -79,7 +79,25 @@
     - post_upgrade_release | length > 0
   changed_when: true
 
-# TODO: Unset rhui release.
+- name: leapp-post-upgrade | Unset release via yum variable for RHUI upgrades
+  ansible.builtin.file:
+    path: /etc/yum/vars/releasever
+    state: absent
+  when:
+    - leapp_upgrade_type == 'rhui'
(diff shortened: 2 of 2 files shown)


changelogs/fragments/unset-rhui-release.yml
```diff
@@ -0,0 +1,5 @@
+---
+major_changes:
```
djdanielsson, line 2: I don't see this as a major change

bparry02: @djdanielsson thanks for taking a look. I listed it as a major change since it will change the behavior for the RHUI upgrade type. Before this change, the version lock was not unset, and after this change it will be. But if the team wants to change it to a minor_change I'm fine with that.

swapdisk (approved):
LGTM

PR #243, opened 2025-03-27
Build changelog first

Tompage1994:
As part of a change in requirements for the release of validated colelctions, we need to ensure the updated changelog is a part of the built collection. This update mirrors https://github.com/redhat-cop/ansible_collections_tooling/pull/59 which was updated to fix this.


djdanielsson (approved):
thanks for this fix

PR #244, opened 2025-03-27
fixing links

djdanielsson:
modified:   .gitignore
	modified:   README.md
	modified:   roles/analysis/README.md
	modified:   roles/remediate/README.md
	modified:   roles/upgrade/README.md

### DIFF
--- a/.gitignore
+++ b/.gitignore
@@ -134,3 +134,4 @@ dmypy.json
 
 # Local collection testing builds
 *.tar.gz
+.ansible/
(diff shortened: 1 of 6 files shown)


djdanielsson:
They would always point back to the main repo/branch yes but without that the readme links on console won't work I believe. I don't see it as a big deal myself but you could see if there is any other way but this is how I know how to do it

swapdisk (approved):
LGTM

Issue #239, opened 2025-03-12
Not all 8-9 post upgrade steps are covered by infra.leapp


jeffmcutter:
I was upgrading Satellite in my lab today to 6.16 and from RHEL 8 to RHEL 9 and I noted some unfamiliar steps in the 8-9 upgrade documentation:

https://docs.redhat.com/en/documentation/red_hat_enterprise_linux/9/html/upgrading_from_rhel_8_to_rhel_9/performing-post-upgrade-tasks-on-the-rhel-9-system_upgrading-from-rhel-8-to-rhel-9

Particularly step 5 regarding rescue kernel.

We might want to review the current upgrade documentation for each 7-8, 8-9, and ultimately 9-10 to ensure full coverage to the current documentation.  I was focused mostly on 6-7 and 7-8 during the initial development.  I'd be interested to help with this if we agree.

rjo-uk:
It's entirely possible I've missed something obvious here - but do we perform the "Remove the old kernel packages from the RHEL 9 system" step on RHEL 8 to RHEL 9 upgrades?

This code hints we do removal in the RHEL 7 to RHEL 8 use case, but I couldn't see something similar for RHEL 8 to RHEL 9?

https://github.com/redhat-cop/infra.leapp/blob/b0054e899918cf4f8393ffdac5f04800a1942ee4/roles/upgrade/tasks/leapp-post-upgrade.yml#L9-L11

jeffmcutter:
@rjo-uk When I initially wrote the collection we were doing 6 -> 7 -> 8, and not 8 -> 9.  I suspect we need to add for 8-9 and 9-10 at some point.  I'm intending to revisit this sometime soon.  Just been trying to get caught up...

jeffmcutter:
> Lines 9 to 11 in [b0054e8](/redhat-cop/infra.leapp/commit/b0054e899918cf4f8393ffdac5f04800a1942ee4)
> 
>  # leapp 7 to 8 documentation cleans up old kernels 
>  - name: leapp-post-upgrade | Block for RHEL 7 kernel cleanup 
>    when: ansible_facts.ansible_local.pre_ripu.distribution_major_version | int == 7

The block is to perform some steps that are only in the 7-8 upgrade.  Step 2 in the docs:

jeffmcutter:
On the overall theme of this issue, the step I encountered in the 8-9 docs is now also in the 7-8 documents step 5.

[...]

PR #250, opened 2025-04-23
Update post upgrade

jeffmcutter:
- Add removal of previous version RHEL packages as the default based upon current RHEL 8 to 9 upgrade documentation.
  - Add update of rescue kernel if an old one is present based upon current RHEL 7 to 8 and 8 to 9 upgrade documentation.
  - Add check for /boot/loader/entries for old RHEL version kernels to post upgrade to RHEL 8+ based upon current RHEL 8 to 9 upgrade documentation.
  - I fixed some markdown issues in the upgrade role README.md file as I was updating the documentation as well.

Addresses https://github.com/redhat-cop/infra.leapp/issues/239.

### DIFF
--- a/roles/upgrade/README.md
+++ b/roles/upgrade/README.md
@@ -31,7 +31,7 @@ Additionally a list of any non-Red Hat RPM packages that were installed on the s
 | kernel_modules_to_unload_before_upgrade | []    | A list of kernel modules to be unloaded prior to running leapp. |
 | post_7_to_8_python_interpreter | /usr/libexec/platform-python | For RHEL 7 to 8 upgrades, /usr/bin/python is discovered but not available post upgrade. For 7 to 8 upgrades, ansible_python_interpreter is set to this value post upgrade reboot prior to reconnecting. |
 | infra_leapp_upgrade_system_roles_collection | fedora.linux_system_roles | Can be one of:<br>- 'fedora.linux_system_roles'<br>- 'redhat.rhel_system_roles' |
-| remove_old_rhel_packages | true                 | Boolean to control whether the previous version RHEL packages should be removed. |
+| remove_old_rhel_packages | true                 | Boolean to control whether the previous version RHEL packages should be removed. Change to false to maintain behavior of infra.leapp prior to  release 1.6.0. |
 
 ## Satellite variables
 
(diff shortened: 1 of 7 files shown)


roles/upgrade/defaults/main.yml
```diff
@@ -158,4 +158,6 @@ post_7_to_8_python_interpreter: /usr/libexec/platform-python
 # - redhat.rhel_system_roles
 infra_leapp_upgrade_system_roles_collection: fedora.linux_system_roles
 
+remove_old_rhel_packages: true
```
djdanielsson, line 161: I wonder if we should default this false because if we do not, people who have been using the collection for a while are likely to pull the update not know about it and then have a very different result vs the prior version.

djdanielsson (changes requested):
I am leaving it as requesting change so we can debate my comment and once decided even if no change I will approve

roles/upgrade/defaults/main.yml
jeffmcutter: That was my initial intent and I know we wanted to have the default behavior stay the same, but it is explicitly called out now in the documentation I don't think having the default be to not follow the documentation is good as new users will likely not realize they should enable it.  Additionally, existing users may well not know and it may be better if they did use it.  Lastly, the validation tasks for 8 to 9 won't pass if the old kernel packages are not removed.

Verification step 4 for 8 to 9:
```
grep -r ".el8" "/boot/loader/entries/" || echo "Everything seems ok."
```

Good for Bob or others to weigh in, but I'm of the opinion we should make it a major change and fix it.

swapdisk: I'm fine with this given we will be releasing as a major change and we want the collection to stay aligned with the Red Hat docs.

roles/upgrade/README.md
```diff
@@ -14,24 +14,24 @@ Additionally a list of any non-Red Hat RPM packages that were installed on the s
...
+| post_upgrade_update     | true                   | Boolean to decide if after the upgrade is performed a dnf update will run. |
+| post_upgrade_unset_release| true                | Boolean used to control whether Leapp's RHSM release lock is unset. |
+| post_upgrade_release    |                       | Optional string used to set a specific RHSM release lock after the Leapp upgrade, but before the final update pass. |
 | kernel_modules_to_unload_before_upgrade | []    | A list of kernel modules to be unloaded prior to running leapp. |
 | post_7_to_8_python_interpreter | /usr/libexec/platform-python | For RHEL 7 to 8 upgrades, /usr/bin/python is discovered but not available post upgrade. For 7 to 8 upgrades, ansible_python_interpreter is set to this value post upgrade reboot prior to reconnecting. |
 | infra_leapp_upgrade_system_roles_collection | fedora.linux_system_roles | Can be one of:<br>- 'fedora.linux_system_roles'<br>- 'redhat.rhel_system_roles' |
-
+| remove_old_rhel_packages | true                 | Boolean to control whether the previous version RHEL packages should be removed. |
```
swapdisk, line 34: Just to make it clear what changed and option to keep it behaving the old way...

```suggestion
| remove_old_rhel_packages | true                 | Boolean to control whether the previous version RHEL packages should be removed. Change to false to maintain behavior of infra.leapp prior to  release 1.6.0. |
```

djdanielsson (approved):
LGTM

swapdisk:
Fixes #239.

PR #256, opened 2025-05-13
updating lint and galaxy files

djdanielsson:


djdanielsson:
trying to decide what to do here... the CI is broken but it appears to be a bug with ansible-compat. so idk if we should change CI test for now to work around it or just ignore it and force merge the change for now and hope they fix it soon

swapdisk:
I'm not a fan of "hope they fix it soon" given it's already been 2 weeks. If it's a trivial workaround to our CI so we don't have to be annoyed anymore, I would vote for that.

swapdisk (approved):
LGTM

PR #259, opened 2025-05-30
trying to do offline lint

djdanielsson:
trying to do offline lint because prior tasks already installs collections


swapdisk (approved):
LGTM.

PR #270, opened 2025-09-18
Ensure leapp packages installed before upgrading

### LINKED ISSUE #267: The upgrade role should install required leapp packages
The typical use case when using this collection is that the `analysis` role is run before the `upgrade` role. However, running only the `upgrade` role is also supported in which case the pre-upgrade report is generated when the role runs the `leapp upgrade` command.

For this use case where the analysis role is not run first, the `check_leapp_analysis_results` boolean must be set to `false` since the report will not yet exist.

This feature was added with #162. 

But I can see there is an issue that the tasks that install the required leapp packages are only in the `analysis` role and not there in the `upgrade` role. They should be in both.

swapdisk:
Fixes #267.

### DIFF
--- a/changelogs/fragments/fix_267.yml
+++ b/changelogs/fragments/fix_267.yml
@@ -1,0 +1,4 @@
+---
+bugfixes:
+  - fixed missing leapp package when upgrade role is run without running the analysis role first
+...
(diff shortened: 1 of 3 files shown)


djdanielsson (approved):
while I think we could use some Jinja to just change which repo if it's 8 or 9 because the packages are the same, it probably isn't worth the small gain you would get vs possible issues

swapdisk:
> while I think we could use some Jinja to just change which repo if it's 8 or 9 because the packages are the same, it probably isn't worth the small gain you would get vs possible issues

I thought about that, but decided better to just mirror how it's done in the analysis role.

PR #286, opened 2025-11-12
fix undefined default variable in the parse_leapp_report role

PeterMocary:
Tests were failing due to an undefined variable in `parse_leapp_report` role.

### DIFF
--- a/roles/parse_leapp_report/defaults/main.yml
+++ b/roles/parse_leapp_report/defaults/main.yml
@@ -2,5 +2,6 @@
 # defaults file for parse_leapp_report
 result_filename_prefix: /var/log/leapp/leapp-report
 result_filename_json: "{{ result_filename_prefix }}.json"
+result_filename_txt: "{{ result_filename_prefix }}.txt"
 result_fact_cacheable: false
 ...
--- a/roles/parse_leapp_report/tasks/main.yml
+++ b/roles/parse_leapp_report/tasks/main.yml
@@ -5,7 +5,7 @@
 
 - name: Collect human readable report results
   ansible.builtin.slurp:
-    src: "{{ result_filename }}"
+    src: "{{ result_filename_txt }}"
   register: results_txt
 
 - name: Collect JSON report results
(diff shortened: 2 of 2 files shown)


spetrosi (approved):

roles/parse_leapp_report/defaults/main.yml
```diff
@@ -2,5 +2,6 @@
 # defaults file for parse_leapp_report
 result_filename_prefix: /var/log/leapp/leapp-report
 result_filename_json: "{{ result_filename_prefix }}.json"
+result_filename_txt: "{{ result_filename_prefix }}.txt"
```
spetrosi: role/defaults/main.yml file is for storing variables that we want user to be able to overwrite - those variables that we document in README and that users will provide when running the role. (Or will not provide, in this case variables get the documented default values).

Here, I guess the aim of the parse_leapp_report role is to make users set the variable `result_filename` when running the role, so that the role can process it.

Now it's confusing, I'd say the role needs to do the following:

1. In roles/parse_leapp_report/defaults/main.yml, define a default path for Leapp pre-upgrade report. Currently, would be something like `parse_leapp_report_path: /var/log/leapp/`.
2. In roles/parse_leapp_report/README.md, document 
`parse_leapp_report_path variable` variable and remove the result_* varialbes from there.
3. In roles/parse_leapp_report/vars/main.yml, define `__parse_leapp_report_json: "{{ parse_leapp_report_path }}/leapp-report.json"` and `__parse_leapp_report_txt: "{{ parse_leapp_report_path }}/leapp-report.txt`.

This, assuming that leapp stores a report to a configurable directory always as leapp-report.txt and leapp-report.json.
If the name (not only path) of report file can be configurable too, the role must have variables for this too.

Current variable names are confusing, and from your changes I see that the role is unusable, because result_filename is not defined anywhere but then used everywhere in the role.

`result_filename` is defined in other roles in  https://github.com/redhat-cop/infra.leapp/blob/main/roles/analysis/vars/main.yml#L5 and https://github.com/redhat-cop/infra.leapp/blob/main/roles/upgrade/vars/main.yml#L5. Maybe the parse_leapp_report expects the variables to be read from there?

spetrosi (changes requested):

roles/parse_leapp_report/defaults/main.yml
spetrosi: @PeterMocary for this PR, let's do a smaller change - your initial issue is that tests fail with `parse_leapp_report` undefined - let's fix this by defining parse_leapp_report from tests. It's expected that this var is undefined because the role expects it to be set.

richm: > There is a simpler way - add tasks files under common/tasks/parse_leapp_report.yml, and then call these tasks like that:

I'm working on this now.

richm: https://github.com/redhat-cop/infra.leapp/pull/290

PR #287, opened 2025-11-12
fix extraction of inhibitor titles during tests

PeterMocary:
Tests were failing when no inhibitors were found on a task that is extracting inhibitor titles.

Jira: RHELMISC-20986


PeterMocary:
Rebased onto main

spetrosi (approved):

PR #290, opened 2025-11-13
refactor: convert parse_leapp_report to a tasks file in common

richm:
Convert parse_leapp_report from being a separate role to being a
tasks file inside of the common role.

Add documentation to the common role README.md

Remove both `common` and `parse_leapp_report` from the main README as these
are not part of the public API.

This "fixes" the undefined `result_filename` variable problem by removing
instances of where it could be called with an undefined value.

catch known upgrade problems in the test so that we do not report false
positives

check for log file presence in handler and skip if not present

Ensure tests do log file cleanup so that subsequent tests will not fail


richm:
[citest]

spetrosi (approved):

PR #295, opened 2025-11-18
refactor: fix code for ansible-lint rather than skip checks

richm:
For certified collections, the Automation Hub gatekeepers prefer to have
as few ansible-lint skips as possible.


richm:
[citest]

PR #297, opened 2025-11-21
Fix upgrade from RHEL 8 to 9

spetrosi:
* Remove outdated (pre-upgrade) bootloader kernel entries. Previously, these were removed only by removing associated RPMs, this change finds kernels that remain and removes them directly with `grubby`.
This could be a bug in systemd-udev, usually kernels remain after removing associated RPMs. To be investigated and reported, but until this is fixed in RHEL, the role should do the workaround applied.
* Do not output information about included custom repositories to avoid exposing URLs
* Remove the workaround to reinstall linux-firmware, it was caused by incorrectly set up repositories. Instead of using the latest z-stream repository for RHEL 9, the CI was using outdated RHEL-9.6.0 repositories.

### DIFF
--- a/roles/upgrade/tasks/leapp-post-upgrade.yml
+++ b/roles/upgrade/tasks/leapp-post-upgrade.yml
@@ -41,6 +41,33 @@
         state: absent
       loop: "{{ old_kernels.files }}"
 
+# Grep for title matching RHEL, then get kernel path from 4 lines above
+- name: >-
+    leapp-post-upgrade | Find kernel paths for outdated RHEL
+    {{ ansible_facts.ansible_local.pre_ripu.distribution_major_version }}
+  ansible.builtin.shell: >-
+    set -o pipefail;
+    export PATH={{ os_path }};
+    grubby --info=ALL | grep -E "^title=.* {{
+    ansible_facts.ansible_local.pre_ripu.distribution_major_version
+    }}\.[0-9]+ " -B 4 | grep "^kernel"
+    | sed -n 's/.*kernel="\([^"]*\)".*/\1/p' || echo ""
+  register: leapp_upgrade_outdated_kernels
+  when: remove_old_rhel_packages
+  changed_when: false
+
+- name: leapp-post-upgrade | Debug kernel paths for outdated RHEL
+  ansible.builtin.debug:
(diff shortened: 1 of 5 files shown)


spetrosi:
[citest]

roles/upgrade/tasks/leapp-post-upgrade.yml
```diff
@@ -146,31 +146,25 @@
...
-# This fails on RHEL 8 to 9 if old kernel packages are not removed.
-# For RHEL 7 to 8, in my testing, no RHEL 7 kernels remain after leapp.
-- name: leapp-post-upgrade | No previous RHEL kernels are present in /boot/loader/entry files
-  ansible.builtin.shell: |
-    set -o pipefail;
-    export PATH={{ os_path }};
-    grep -r ".el{{ ansible_facts.ansible_local.pre_ripu.distribution_major_version }}" "/boot/loader/entries/" || echo "Everything seems ok."
-  register: boot_loader_entries
+- name: leapp-post-upgrade | Remove outdated RHEL kernels from bootloader using grubby
+  ansible.builtin.command: grubby --remove-kernel="{{ item }}"
```
richm:
```suggestion
  ansible.builtin.command: grubby --remove-kernel={{ item | quote }}
```
Better to use quote instead of double quotes

roles/upgrade/tasks/leapp-post-upgrade.yml
```diff
@@ -146,31 +146,25 @@
...
+  loop: "{{ leapp_upgrade_outdated_kernels.stdout_lines }}"
   when:
-    - remove_old_rhel_packages # This will fail if the old kernel packages weren't removed.
-    - ansible_facts.ansible_local.pre_ripu.distribution_major_version | int >= 8
-  changed_when: false
-  failed_when: boot_loader_entries.stdout != 'Everything seems ok.'
+    - remove_old_rhel_packages
+    - leapp_upgrade_outdated_kernels.stdout| length > 0
```
richm:
```suggestion
    - leapp_upgrade_outdated_kernels.stdout | length > 0
```

spetrosi:
[citest]

spetrosi:
[citest]

spetrosi:
[citest]

spetrosi:
[citest]

PR #300, opened 2025-11-26
refactor: Refactor remediations - extract common code into role main - do not use fail/rescue

richm:
There was a lot of code duplication in the remedation task files.  This has been extracted
into the remediation role main task file.

The remediation tasks were using fail/rescue for flow control, which makes looking at the
logs very confusing, as you really have to know which tasks to exclude when looking for errors.
Instead, use plain old conditionals with blocks for tasks to run conditionally.  NOTE: The
old way of using `rescue` would essentially catch and ignore all errors - this seems
wrong to me, but perhaps there is a reason.  This means that now, remediation may raise
errors where before it would not.

Use ansible good practices:

* Use `vars` instead of `set_fact` and keep the variables set at the lowest necessary scope
* Use filters instead of loops
* Use conditional evaluation instead of comparing a value `== true`
* Use tests such as `is success` rather than looking at the register rc value directly

### DIFF
--- a/roles/remediate/tasks/leapp_cgroups-v1_enabled.yml
+++ b/roles/remediate/tasks/leapp_cgroups-v1_enabled.yml
@@ -1,71 +1,42 @@
 ---
+- name: leapp_cgroups-v1_enabled | Continue when leapp report is missing
+  ansible.builtin.debug:
+    msg: "Leapp report missing. Skipping this task."
+  when: leapp_report_missing | default(false)
+
 - name: leapp_cgroups-v1_enabled | Disable cgroups-v1 in favor of the new version cgroups-v2
   vars:
-    leapp_report_location: /var/log/leapp/leapp-report.json
     leapp_inhibitor_title: cgroups-v1 enabled on the system
+    remediation_matches: "{{ leapp_report_data.entries | selectattr('title', 'match', leapp_inhibitor_title) |
+      map(attribute='detail.remediations') | flatten | list }}"
+    leapp_inhibitor_remediation: "{{ remediation_matches | selectattr('type', 'eq', 'command') | first
+      if remediation_matches | length > 0 else {} }}"
   when:
+    - not leapp_report_missing | default(false)
     - ansible_distribution == 'RedHat'
-    - ansible_distribution_major_version|int == 9
+    - ansible_distribution_major_version | int == 9
   block:
-    - name: leapp_cgroups-v1_enabled | Check that the leapp-report.json exists
-      ansible.builtin.stat:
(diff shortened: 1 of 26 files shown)


richm:
[citest]

amott-rh:
@richm Probably in general RHEL 6 has become less relevant, but I wrote that remediation precisely because I had customer engagements that were upgrading RHEL 6 servers and ran into the /usr issue. I also have colleagues in the same position, but agree it's generally rare.

roles/remediate/tasks/leapp_cgroups-v1_enabled.yml
```diff
@@ -1,71 +1,42 @@
...
-      when: leapp_report_missing | d(false)
+        msg: "No matching entry found in leapp report. Skipping this task."
+      when: leapp_inhibitor_remediation | length == 0
+
+    - name: leapp_cgroups-v1_enabled | Process remediation
+      when: leapp_inhibitor_remediation | length > 0
+      block:
+        - name: leapp_cgroups-v1_enabled | Show the remediation command
+          ansible.builtin.debug:
+            msg: "{{ leapp_inhibitor_remediation.context | join(' ') }}"
```
spetrosi:
```suggestion
            bar: leapp_inhibitor_remediation.context | join(' ')
```

roles/remediate/tasks/leapp_relative_symlinks.yml
```diff
@@ -1,54 +1,38 @@
...
+    - name: leapp_relative_symlinks | Process remediation
+      when: remediation | length > 0
+      block:
+        - name: leapp_relative_symlinks | Output command to be executed
+          ansible.builtin.debug:
+            msg: "{{ remediation.context }}"
+
+        # The old version passed the list directly to the command.  I didn't think
+        # this was possible.  Guard against something weird here.  Assume non-string
+        # is a list.
```
spetrosi: @PeterMocary do you know where leapp inhibitors are stored? Maybe there is some KCS about this inhibitor that can provide us more info

richm:
[citest]

richm:
Any objection to merging this?  We will be adding tests for remediations in upcoming PRs

spetrosi (approved):

amott-rh:
@richm It was a while ago, but I probably used a RHEL 7 machine running Ansible 2.9. I was working with a customer who needs to migrate a good number of RHEL 6 machines to RHEL 7, then 8 then 9, and these playbooks were the most reliable and reproducible way to achieve this. As I understand it, this project is still in progress, but I absolutely understand your point - there comes a time when old version can't be reasonably supported by them.

PR #302, opened 2025-12-01
Fix 9 to 10 upgrade tests

PeterMocary:
The post-upgrade task which updates the system packages failed when
running upgrade with custom repositories. Since leapp does not update the
repositories in `/etc/yum.repos.d` when upgrading with custom repositories,
the post-upgrade update task is now skipped in the case when 
the `leapp_upgrade_type` is custom and when the `local_repos_post_upgrade`
 are not defined.

~This PR includes commits from the PR#285 already since these fix some issues
with python interpreter encountered in 9 to 10 tests.~ The last commit acts as 
a minor fix to the clean up of role tests, since executing these without it caused
 them to fail. The analysis role test fails on leapp being not installed since its
 dependencies are not satisfied and there is  no rescue block for error handling
 as in the upgrade role test. Currently these are here just to execute the roles
 independently. We need to decide what to do with them in future.

### DIFF
--- a/roles/analysis/tests/tests_default.yml
+++ b/roles/analysis/tests/tests_default.yml
@@ -3,6 +3,21 @@
   hosts: all
   vars:
     job_name: test_analysis
-  roles:
-    - analysis
+  tasks:
+    - name: Test | Run analysis
+      block:
+        - name: Test | Run role analysis
+          ansible.builtin.include_role:
+            name: infra.leapp.analysis
+      always:
+        - name: Cleanup | Remove log files
+          tags: tests::cleanup
+          ansible.builtin.shell:
+            executable: /bin/bash
+            cmd: |
+              set -euxo pipefail
+              rm -f /var/log/leapp/leapp-upgrade.log
+              rm -f /var/log/ripu/ripu.log*
+          changed_when: true
+
 ...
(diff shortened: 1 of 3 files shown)


roles/upgrade/README.md
```diff
@@ -29,7 +29,7 @@ Additionally a list of any non-Red Hat RPM packages that were installed on the s
 | post_upgrade_unset_release| true                | Boolean used to control whether Leapp's RHSM release lock is unset. |
 | post_upgrade_release    |                       | Optional string used to set a specific RHSM release lock after the Leapp upgrade, but before the final update pass. |
 | kernel_modules_to_unload_before_upgrade | []    | A list of kernel modules to be unloaded prior to running leapp. |
-| post_7_to_8_python_interpreter | /usr/libexec/platform-python | For RHEL 7 to 8 upgrades, /usr/bin/python is discovered but not available post upgrade. For 7 to 8 upgrades, ansible_python_interpreter is set to this value post upgrade reboot prior to reconnecting. |
```
richm: You should add a note that `post_7_to_8_python_interpreter` is Deprecated, use `post_upgrade_python_interpreter` instead.

spetrosi:
[citest]

spetrosi:
[citest]

PR #312, opened 2025-12-11
Remove dependency on community.general.modprobe module

PeterMocary:
In order to test it simply add the following into the execution (I've tried 8to9 already)

```
kernel_modules_to_unload_before_upgrade:
      - "joydev"
      - "ahci"
```
Jira: RHEL-134414

### DIFF
--- a/roles/upgrade/tasks/remove-kernel-modules.yml
+++ b/roles/upgrade/tasks/remove-kernel-modules.yml
@@ -1,0 +1,16 @@
+---
+- name: remove-kernel-modules | Get names of loaded kernel modules
+  ansible.builtin.command:
+    cmd: awk '{print $1}' /proc/modules
+  register: loaded_kernel_modules
+  changed_when: false
+
+- name: remove-kernel-modules | Remove specified kernel modules
+  ansible.builtin.command:
+    cmd: modprobe -r {{ kernel_module }}
+  changed_when: true
+  loop: "{{ __kernel_modules_to_remove | list }}"
+  loop_control:
+    loop_var: kernel_module
+  when: kernel_module in loaded_kernel_modules.stdout_lines
+...
--- a/.ansible-lint
+++ b/.ansible-lint
@@ -27,7 +27,6 @@ exclude_paths:
 # not being able to pull community.general (not being validated)
 mock_modules:
   - community.general.redhat_subscription
-  - community.general.modprobe
   - community.general.archive
   - community.general.yum_versionlock
   - community.general.rhsm_repository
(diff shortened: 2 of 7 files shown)


roles/upgrade/tasks/remove-kernel-modules.yml
```diff
@@ -0,0 +1,16 @@
+---
+- name: remove-kernel-modules | Get names of loaded kernel modules
+  ansible.builtin.command:
+    cmd: awk '{print $1}' /proc/modules
+  register: loaded_kernel_modules
+  changed_when: false
+
+- name: remove-kernel-modules | Remove kernel module {{ kernel_module }}
```
spetrosi: Variables are not read at the time that task starts, so they are not displayed in the title. This is displayed as plain test `TASK [infra.leapp.upgrade : remove-kernel-modules | Remove kernel module {{ kernel_module }}] ***`. So remove this var.

spetrosi (commented):
I'd like to have this tested first, Rich suggested a good way to test remediations in https://github.com/redhat-cop/infra.leapp/pull/309. We can wait until he implements this in his PR and then add a similar test here.

spetrosi (commented):
Rich has submitted a test for a remediation in #309, can you add a similar test for the remediation that you do here?

spetrosi:
> So, for now, I think we have to keep the modprobe module, and even use it in the kernel module remediation.

@richm the remediation that you mention uses community.general.modprobe module already https://github.com/redhat-cop/infra.leapp/blob/main/roles/remediate/tasks/leapp_loaded_removed_kernel_drivers.yml#L27-L32

spetrosi:
And I agree that "Disable module options in /etc/modprobe.d" is a pain. So let's keep the community.general.modprobe in the upstream, and in the downstream ship it as part of the collection.

Since we have a full collection code here, maybe it makes more sense to ship modprobe.py in the upstream in the first place? This will remove the dependency on community.general making it easier for users to use the collection. And saves some work for us too. @richm WDYT?

richm:
@PeterMocary While trying to fix the CI test failure, I think I inadvertently superseded your PR with mine https://github.com/redhat-cop/infra.leapp/pull/318 - @spetrosi wdyt?

spetrosi:
https://github.com/redhat-cop/infra.leapp/pull/318 uses community.general.modprobe extensively, so it's clear that we want to keep it, so this PR can be closed.
But we need to decide how to handle community.general.modprobe. How about my suggestion in https://github.com/redhat-cop/infra.leapp/pull/312#issuecomment-3669908838?

PeterMocary:
[citest]

PR #317, opened 2025-12-17
Prefix variables from defaults/main.yml with `leapp_` for consistency

spetrosi:
1. Prefix variables that are not prefixed already.
2. Edit variables that had leapp in names but not as prefix to use leapp_ as prefix.
3. Assign old variables to new varnames if an old name is set for backward compatibility.
4. Save variables with normal values (without the default filter for backward compatilibility) in a comment so that we can easily remove compatibility in future when needed. Say, in couple years when we are sure that no users have old deprecated variables set.

### DIFF
--- a/roles/common/defaults/main.yml
+++ b/roles/common/defaults/main.yml
@@ -1,7 +1,6 @@
 ---
 # defaults file for common
-os_path: $PATH
-log_directory: /var/log/ripu
-log_file: "{{ log_directory }}/ripu.log"
-
+leapp_os_path: $PATH
+leapp_log_directory: /var/log/ripu
+leapp_log_file: "{{ leapp_log_directory }}/ripu.log"
 ...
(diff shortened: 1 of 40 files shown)


roles/common/defaults/main.yml
```diff
@@ -1,7 +1,6 @@
 ---
 # defaults file for common
-os_path: $PATH
-log_directory: /var/log/ripu
-log_file: "{{ log_directory }}/ripu.log"
-
+leapp_os_path: $PATH
```
richm:
```suggestion
leapp_os_path: "{{ os_path | default('$PATH') }}"
```
?  Unless I'm missing something?

spetrosi: Common role is used from within other roles only so it doesn't require to be pre-populated.

richm:
I'm working on debugging the el7 test failure - go ahead and merge this PR

PR #318, opened 2025-12-17
fix: ensure removed kernel modules are not loaded at boot; add module remediation test

richm:
Ensure that removed/unloaded kernel modules are not reloaded at boot.  It seems that
the upgrade run can reboot the system after analysis but before upgrade, and if the
unsupported kernel modules are not disabled, they will be loaded again at boot.

I have also added a test for the unsupported kernel module remediation.

This also refactors the kernel module load/unload code into a new common
tasks file manage_kernel_modules.yml which implements the modprobe plus the
module file handling.

### DIFF
--- a/roles/upgrade/tasks/rmmod-kernel-modules.yml
+++ b/roles/upgrade/tasks/rmmod-kernel-modules.yml
@@ -1,7 +1,0 @@
----
-- name: rmmod-kernel-modules | Unload module {{ item }}
-  community.general.modprobe:
-    name: "{{ item }}"
-    state: absent
-
-...
--- a/tests/tasks/setup/remediate_removed_kernel_drivers.yml
+++ b/tests/tasks/setup/remediate_removed_kernel_drivers.yml
@@ -1,0 +1,29 @@
+# set up the test system to test removed kernel drivers remediation
+# see KERNEL_MODULES.md for more details
+---
+- name: setup | remediate_removed_kernel_drivers | Set list of test kernel modules
+  ansible.builtin.set_fact:
+    leapp_test_kernel_modules: "{{ __test_modules.get(__platform, []) }}"
+  vars:
+    __platform: "{{ ansible_facts['os_family'] }}_{{ ansible_facts['distribution_major_version'] }}"
+    __test_modules:
+      RedHat_7:
+        - 3w-9xxx
+        - pata_acpi
+        - tulip
+      RedHat_8:
+        - dnet
+        - dlci
+        - liquidio
+      RedHat_9:
+        - qla3xxx
+        - vmw_pvrdma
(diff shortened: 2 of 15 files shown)


richm:
[citest]

richm:
[citest]

roles/upgrade/tasks/rmmod-kernel-modules.yml
```diff
@@ -3,5 +3,6 @@
   community.general.modprobe:
     name: "{{ item }}"
     state: absent
+    persistent: absent
```
spetrosi: That's why remediation failed?

richm: Actually, `persistent: absent` doesn't work unless the module is already in a modprobe.d file.  I think the modprobe module is not sufficient for our use.  I'm working on a different way to do this.

richm:
[citest]

richm:
[citest]

tests/tasks/verify/remediate_removed_kernel_drivers.yml
```diff
@@ -0,0 +1,11 @@
...
+---
+- name: verify | remediate_removed_kernel_drivers | Verify the removed kernel drivers remediation
+  ansible.builtin.shell: set -o pipefail; lsmod | awk '{print $1}'
+  changed_when: false
+  register: lsmod_output
+  vars:
+    modlist: "{{ lsmod_output.stdout_lines | map('trim') | list }}"
+    diff: "{{ modlist | intersect(leapp_test_kernel_modules) | list }}"
+  when: leapp_test_kernel_modules | default([]) | length > 0
+  failed_when: diff | length > 0
```
PeterMocary: Missing the `...` at the end

tests/tasks/setup/remediate_removed_kernel_drivers.yml
```diff
@@ -0,0 +1,18 @@
...
+    leapp_test_kernel_modules:
+      - pata_acpi
+
+- name: setup | remediate_removed_kernel_drivers | Load the test kernel modules
+  ansible.builtin.include_role:
+    name: infra.leapp.common
+    tasks_from: manage_kernel_modules.yml
+  vars:
+    leapp_kernel_modules: "{{ leapp_test_kernel_modules }}"
+    leapp_kernel_modules_state: present
```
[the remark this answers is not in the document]
spetrosi: We got rid of this rule, so not needed for new files. This is a yaml standard, but it doesn't help in any way.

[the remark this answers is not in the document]
richm: The `...` isn't required by Ansible.

Issue #320, opened 2025-12-28
template error while templating string: no filter named 'split'. String: {{ results_txt.content | b64decode | split('\n') }}


stu93303148:
file_name：infra/leapp/roles/parse_leapp_report/tasks/main.yml

~~~
 15
 16 - name: Parse report results
 17   ansible.builtin.set_fact:
 18     cacheable: "{{ result_fact_cacheable }}"
 19     leapp_report_txt: "{{ results_txt.content | b64decode | split('\n') }}"
 20    leapp_report_json: "{{ results_json.content | b64decode | from_json }}"
 21
~~~

Error msg

```
{
  "msg": "template error while templating string: no filter named 'split'. String: {{ results_txt.content | b64decode | split('\n') }}",
  "_ansible_no_log": false
}
```

On line 19, should we change it to ``` leapp_report_txt: "{{ (results_txt.content | b64decode).split('\n') }}" ``` ?

richm:
@stu93303148 there is a built-in `split` filter added in 2.11 - https://docs.ansible.com/projects/ansible/latest/playbook_guide/playbooks_filters.html#transforming-strings-into-lists

What version of ansible are you using?

PR #324, opened 2025-12-29
fix: use split method instead of filter for 2.9 support

richm:
The `split` filter was added in 2.11, so we cannot use it in Ansible 2.9.
Instead, use the `split` string method.


richm:
fixes https://github.com/redhat-cop/infra.leapp/issues/320

richm:
[citest]

PR #343, opened 2026-01-20
fix: Vendor infra.leapp.modprobe inside the collection

spetrosi:
This is required to remove dependency on community.general collection.

I've done this by downloading community-general-11.4.2 from https://galaxy.ansible.com/download/community-general-11.4.2.tar.gz and copying `plugins/modules/modprobe.py` from it to this collection.
I took version 11.4.2 because this is what rhel-system-roles uses, this version is verified to work with all Python versions that we need to support.
community.general now has a tendency to add features that are not supported on older Python which RHEL 7 uses, so we cannot use latest version.
In the module I've added a warning `  - "WARNING: Do not use this plugin directly! It is only for role internal use."` and replaced `community.general.` with `infra.leapp.`.
Then had to copy plugins/doc_fragments/attributes.py because it is needed to build module documentation.

### DIFF
--- a/.ansible-lint
+++ b/.ansible-lint
@@ -27,9 +27,6 @@ exclude_paths:
 # not being able to pull community.general (not being validated)
 mock_modules:
   - community.general.redhat_subscription
-  - community.general.modprobe
-  - community.general.archive
-  - community.general.rhsm_repository
   - fedora.linux_system_roles.selinux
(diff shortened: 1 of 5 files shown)


spetrosi:
@PeterMocary @tomasfratrik FYI

richm (approved):

PR #349, opened 2026-02-05
fix satellite subscription handling

PeterMocary:
As a followup on PR #348, fix the satellite subscription problem which originates from use of handlers and `include_role` incompatibility. Since using `include_role` is not supported in handlers (https://github.com/ansible/ansible/issues/20493), it is not a good idea to wrap it in `include_tasks` and use it anyways. Thus, this patch removes the handlers entirely and converts them into regular tasks. This solution is inline with the original idea (https://github.com/redhat-cop/infra.leapp/pull/348#issuecomment-3833091991) and basically similar to the key handling in the upgrade role.

Not tested yet, although I'm already able to setup satellite host and upgrade it. I'll try it soon.

### DIFF
--- a/roles/analysis/README.md
+++ b/roles/analysis/README.md
@@ -14,13 +14,13 @@ This role will not fail if there are inhibitors found, it will throw a warning. 
 
 ## Satellite variables
 
-Activation keys provide a method to identify content views available from Red Hat Satellite. To do in-place upgrades using Satellite, a content view including both the current RHEL version and the next version must be created. Use these variables to specify the activation keys for the required content views.
+Activation keys provide a method to identify content views available from Red Hat Satellite. To do in-place upgrades using Satellite, both the current RHEL version and the next RHEL version repositories must be available. Use these variables to specify the activation keys for the required content views. In case the system uses a different content view than the one used for the upgrade, one can specify the `leapp_satellite_activation_key_pre_leapp` in order to re-register to it after the analysis concludes, leaving the system in its original state. If not specified, the system will remain registered to the `leapp_satellite_activation_key_leapp` used during the analysis.
 
 | Name                  | Type | Default value           | Description                                     |
 |-----------------------|------|-------------------------|-------------------------------------------------|
 | leapp_satellite_organization  | String   | "" | Organization used in Satellite definition |
-| leapp_satellite_activation_key_pre_leapp | String | "" | Activation key for the current RHEL version content view to re-register to after analysis |
 | leapp_satellite_activation_key_leapp     | String | "" | Activation key for the content view including both the current RHEL version and the next version |
+| leapp_satellite_activation_key_pre_leapp | String | leapp_satellite_activation_key_leapp | Activation key for the current RHEL version content view to re-register to after analysis |
 | leapp_repos_enabled    | List | [] | Satellite repo for the satellite client RPM install |
 
 ## Custom repos variables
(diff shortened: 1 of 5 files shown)


richm:
[citest]

PeterMocary:
[citest]

roles/upgrade/tasks/redhat-upgrade-tool-upgrade-el6.yml
```diff
@@ -22,22 +22,15 @@
...
-    force_register: true
-  register: sub_man_reg
+  ansible.builtin.include_role:
+    name: infra.leapp.common
+    tasks_from: register-satellite-el6.yml
+  vars:
+    satellite_organization: "{{ leapp_satellite_organization }}"
+    satellite_activation_key: "{{ leapp_satellite_activation_key_leapp }}"
   when:
     - leapp_upgrade_type == 'satellite'
```
spetrosi:
```suggestion
when: leapp_upgrade_type == 'satellite'
```

spetrosi (approved):

PeterMocary:
[citest]

roles/analysis/tasks/analysis-leapp.yml
```diff
@@ -92,6 +91,19 @@
...
+- name: analysis-leapp | Restore original Satellite activation key
+  ansible.builtin.include_role:
+    name: infra.leapp.common
+    tasks_from: register-satellite.yml
+  vars:
+    satellite_organization: "{{ leapp_satellite_organization }}"
+    satellite_activation_key: "{{ leapp_satellite_activation_key_pre_leapp }}"
+  when:
+    - leapp_upgrade_type == 'satellite'
+    - leapp_satellite_activation_key_pre_leapp != leapp_satellite_activation_key_leapp
```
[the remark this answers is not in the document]
spetrosi: In my test I set only leapp_satellite_activation_key_leapp.
And both these conditions passed and the included tasks got executed and failed because in reality `leapp_satellite_activation_key_pre_leapp=""`. 
So there should be one more check `- leapp_satellite_activation_key_pre_leapp | length > 0`.

roles/analysis/README.md
```diff
@@ -14,13 +14,13 @@ This role will not fail if there are inhibitors found, it will throw a warning.
...
 
 | Name                  | Type | Default value           | Description                                     |
 |-----------------------|------|-------------------------|-------------------------------------------------|
-| leapp_satellite_organization  | String   | null | Organization used in Satellite definition |
-| leapp_satellite_activation_key_pre_leapp | String | null | Activation key for the current RHEL version content view |
-| leapp_satellite_activation_key_leapp     | String | null | Activation key for the content view including both the current RHEL version and the next version |
+| leapp_satellite_organization  | String   | "" | Organization used in Satellite definition |
+| leapp_satellite_activation_key_leapp     | String | "" | Activation key for the content view including both the current RHEL version and the next version |
+| leapp_satellite_activation_key_pre_leapp | String | leapp_satellite_activation_key_leapp | Activation key for the current RHEL version content view to re-register to after analysis |
```
spetrosi: I have a last concern about variables, these are the only variables that has _leapp at the beginning AND at the end. Can you rename them to be this: 

leapp_satellite_activation_key_leapp - used during analysis and upgrade, is the same key for both cases so can be just `leapp_satellite_activation_key`.
leapp_satellite_activation_key_pre_leapp - used to re-register to after analysis. Can be `leapp_satellite_activation_key_post_analysis`.
leapp_satellite_activation_key_post_leapp - used to register to after the upgrade. Can be `leapp_satellite_activation_key_post_upgrade`.

This would bring some consistency. This wording with pre and post feels confusing to me? What do you think?
Can be done in a next PR if you wish.

PR #353, opened 2026-02-12
fix: Ensure leapp_infra_upgrade_system_roles_collection is used

spetrosi:
* Some parts were using plain-text collection for system roles, fixing it.
* Removing extra collection dependencies from readmes and collection-requirements.yml files.
* Add leapp_infra_upgrade_system_roles_collection to roles where it's applicable. Both to README and defaults/main.yml
* Remove roles/upgrade/meta/argument_spec.yml because it listed only a single outdated variable.
* Remove roles/upgrade/meta/collection-requirements.yml because meta/collection-requirements.yml should be on the collection root level only.
* Reminder to self to make the following changes in RPM spec:
    * Replace the value of leapp_infra_upgrade_system_roles_collection with redhat.rhel_system_roles
    * Replace documentation of the default value ofleapp_infra_upgrade_system_roles_collection in roles READMEs.

### DIFF
--- a/roles/analysis/defaults/main.yml
+++ b/roles/analysis/defaults/main.yml
@@ -184,5 +184,6 @@ leapp_bypass_fs_checks: false
 # For community/upstream, use 'fedora.linux_system_roles'.
 # For the RHEL, AAP, use 'redhat.rhel_system_roles'
 # infra_leapp_upgrade_system_roles_collection is deprecated, use leapp_infra_upgrade_system_roles_collection instead
+# leapp_infra_upgrade_system_roles_collection: fedora.linux_system_roles
 leapp_infra_upgrade_system_roles_collection: "{{ infra_leapp_upgrade_system_roles_collection | default('fedora.linux_system_roles') }}"
 ...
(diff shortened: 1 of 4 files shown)


roles/analysis/defaults/main.yml
```diff
@@ -180,12 +180,9 @@ leapp_async_poll_interval: "{{ async_poll_interval | default(60) }}"
...
-# Set which Ansible Collection to use for the Linux System Roles.
-# For community/upstream, use 'fedora.linux_system_roles'
-# For the RHEL System Roles, use 'redhat.rhel_system_roles'
-# - fedora.linux_system_roles
-# - redhat.rhel_system_roles
+# Set which Ansible Collection to use for System Roles.
+# For community/upstream, use 'fedora.linux_system_roles'.
+# For the RHEL, AAP, use 'redhat.rhel_system_roles'
 # infra_leapp_upgrade_system_roles_collection is deprecated, use leapp_infra_upgrade_system_roles_collection instead
-# leapp_infra_upgrade_system_roles_collection: fedora.linux_system_roles
```
PeterMocary: Same here, this line should stay.

spetrosi:
I want to rename `leapp_infra_upgrade_system_roles_collection` to a shorter and preciser `leapp_system_roles_collection`. Will push a new PR after this is merged.

PR #359, opened 2026-02-16
rename variables for satellite activation keys

PeterMocary:
Renamed variables that specify activation keys used by the collection to register the system during different phases of the upgrade.

`leapp_satellite_activation_key_leapp` -> `leapp_satellite_activation_key`
`leapp_satellite_activation_key_pre_leapp` -> `leapp_satellite_activation_key_post_analysis` `leapp_satellite_activation_key_post_leapp` -> `leapp_satellite_activation_key_post_upgrade`

This patch also updates the readmes and example playbooks relevant to this change and fixes the default value of `leapp_upgrade_type` variable in the readme as well.


changelogs/fragments/359-rename-satellite-activation-key-variables.yml
```diff
@@ -0,0 +1,4 @@
+---
+breaking_changes:
```
richm: It doesn't need to be a breaking change.  You can make the new variable default to the value of the old variable if it is set.  see https://github.com/redhat-cop/infra.leapp/pull/317/changes e.g. in the defaults/main.yml

```
leapp_new_variable: "{{ leapp_old_variable | default(default value) }}"
```

e.g.

```
leapp_satellite_activation_key: "{{ leapp_satellite_activation_key_leapp | default('') }}"
leapp_satellite_activation_key_post_analysis: "{{ leapp_satellite_activation_key_pre_leapp | default(leapp_satellite_activation_key) }}"
```

richm (approved):

Issue #355, opened 2026-02-12
Tables with variables in roles are not consistent


spetrosi:
Upgrade and remediations roles tables do not have a `Type` column name, while analysis role have it. Would be good to have consistency.

PR #360, opened 2026-02-16
Update readme tables

PeterMocary:
The format of tables in READMEs that containing collection variables was
inconsistent and hard to read. This patch unifies the tables by adding
the expected type to every variable and makes the types and values
consistent. Better spacing was also added to improve readability in raw
format.

Issue #355.


spetrosi:
`git rebase origin/main -i` allows you to drop old commits from your branch.

spetrosi (approved):

PR #387, opened 2026-04-06
ci: add Ansible partner certification check

richm:
https://github.com/ansible-collections/partner-certification-checker/blob/main/README.md

This checks that the collection can be imported into Automation Hub, among other
things.

### DIFF
--- a/.ansible-lint
+++ b/.ansible-lint
@@ -22,14 +22,6 @@ exclude_paths:
 # strict: true
 # verbosity: 1
 
-# Mock modules or roles in order to pass ansible-playbook --syntax-check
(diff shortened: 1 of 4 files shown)


richm:
the linux_system_roles collection does not have both MANIFEST.json and galaxy.yml - fixing

richm:
> the linux_system_roles collection does not have both MANIFEST.json and galaxy.yml - fixing

No, it was the mock_modules in the .ansible-lint that was causing the problems

swapdisk (approved):
Assuming this works, looks like a good thing. Thanks!

PR #392, opened 2026-05-04
update report titles and keys

PeterMocary:
Some of the titles and keys for reports were updated in the
leapp-repository and this collection needs to reflect it. This patch
simply adds new titles to the title/key maps in order to preserve
functionality with older versions.

The title searched in remediations is updated without preserving the
original one, since the searching by title is a fallback option as
opposed to searching by report key which is the default. Thus, it should
be enough to preserve the old keys there.

The titles and keys were changed in leapp-repository PR#1505, wich
affects only RHEL 8 and above, as RHEL 7 is not maintained in the
upstream repositoy anymore.

Following are relevant **changes in leapp-repository** that are reflected by
this PR.

Titles used before:
- RHEL 7  - Leapp detected loaded kernel drivers which have been removed in RHEL 8. Upgrade cannot proceed.
- RHEL 8  - Leapp detected loaded kernel drivers which have been removed in RHEL 9. Upgrade cannot proceed.
- RHEL 9  - Leapp detected loaded kernel drivers which have been removed in RHEL 10. Upgrade cannot proceed.

Titles used now:
- RHEL 7  - Leapp detected loaded kernel drivers which have been removed in RHEL 8. Upgrade cannot proceed.
- RHEL 8+ - Leapp detected loaded kernel drivers that have been removed in the target OS. Upgrade cannot proceed.

Keys used before:
- RHEL 7  - `f08a07da902958defa4f5c2699fae9ec2eb67c5b`
- RHEL 8  - `7ae2961239eec9905e2580fa6309a589b1dca953`
- RHEL 9  - `48e04852631245aa4afee9adcdc7c2375b8cffb8`

Key used now:
- RHEL 7+ - `f08a07da902958defa4f5c2699fae9ec2eb67c5b`

---

Changes being reflected come from here:
https://github.com/oamg/leapp-repository/pull/1505

This was an update that makes sure titles dont contain distro and its version in the titles since leapp now supports conversions. Notice that the "The installed KDE environment is unavailable on RHEL 8" in the title_map was not changed since it is related to 7to8 upgrade only which is not maintained by leapp upstream anymore.

Jira: RHEL-170715

### DIFF
--- a/roles/analysis/vars/main.yml
+++ b/roles/analysis/vars/main.yml
@@ -54,9 +54,6 @@ __leapp_inhibitors_title_map:
   "Firewalld Configuration AllowZoneDrifting Is Unsupported": leapp_firewalld_allowzonedrifting
   "Firewalld Service tftp-client Is Unsupported": leapp_firewalld_unsupported_tftp_client
   "Legacy network configuration found": leapp_legacy_network_configuration
-  "Leapp detected loaded kernel drivers which have been removed in RHEL 8. Upgrade cannot proceed.": leapp_loaded_removed_kernel_drivers
-  "Leapp detected loaded kernel drivers which have been removed in RHEL 9. Upgrade cannot proceed.": leapp_loaded_removed_kernel_drivers
-  "Leapp detected loaded kernel drivers which have been removed in RHEL 10. Upgrade cannot proceed.": leapp_loaded_removed_kernel_drivers
   "efibootmgr package is required on EFI systems": leapp_missing_efibootmgr
   "Package \"leapp-rhui-aws\" is missing": leapp_missing_pkg
   "Package \"leapp-rhui-google\" is missing": leapp_missing_pkg
@@ -79,6 +76,10 @@ __leapp_inhibitors_title_map:
   "Detected RPMs with RSA/SHA1 signature": leapp_rpms_with_rsa_sha1_detected
   "The installed KDE environment is unavailable on RHEL 8": leapp_unavailable_kde
   "Cannot perform the VDO check of block devices": leapp_vdo_check_needed
+  "Leapp detected loaded kernel drivers which have been removed in RHEL 8. Upgrade cannot proceed.": leapp_loaded_removed_kernel_drivers
+  "Leapp detected loaded kernel drivers which have been removed in RHEL 9. Upgrade cannot proceed.": leapp_loaded_removed_kernel_drivers
+  "Leapp detected loaded kernel drivers which have been removed in RHEL 10. Upgrade cannot proceed.": leapp_loaded_removed_kernel_drivers
+  "Leapp detected loaded kernel drivers that have been removed in the target OS. Upgrade cannot proceed.": leapp_loaded_removed_kernel_drivers
 
 # Maps the key attribute of an inhibitor from a report to a remediation task file name
 __leapp_inhibitors_key_map:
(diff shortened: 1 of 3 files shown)


roles/analysis/vars/main.yml
```diff
@@ -115,5 +115,8 @@ __leapp_inhibitors_key_map:
   f16f40f49c2329a2691c0801b94d31b6b3d4f876: leapp_rpms_with_rsa_sha1_detected
   4bada539189b239213b653dae48c383d20dfe97a: leapp_unavailable_kde
   429a99e13b19a7eebadbb8cb35233d8119bcf255: leapp_vdo_check_needed
+  f08a07da902958defa4f5c2699fae9ec2eb67c5b: leapp_loaded_removed_kernel_drivers
+  7ae2961239eec9905e2580fa6309a589b1dca953: leapp_loaded_removed_kernel_drivers # TODO: Can be removed with the next leapp release (Nov 2026)
```
richm:
```suggestion
  7ae2961239eec9905e2580fa6309a589b1dca953: leapp_loaded_removed_kernel_drivers # TODO: Can be removed with the next leapp release (Nov 2026)
```

roles/remediate/tasks/leapp_loaded_removed_kernel_drivers.yml
```diff
@@ -9,9 +9,10 @@
 
     leapp_inhibitor_keys:
       - f08a07da902958defa4f5c2699fae9ec2eb67c5b
-      - 7ae2961239eec9905e2580fa6309a589b1dca953
+      # TODO: Following keys can be removed with the next leapp release (Nov 2026)
+      - 7ae2961239eec9905e2580fa6309a589b1dca953
```
richm:
```suggestion
      - 7ae2961239eec9905e2580fa6309a589b1dca953
```

roles/analysis/vars/main.yml
```diff
@@ -79,6 +76,12 @@ __leapp_inhibitors_title_map:
   "Detected RPMs with RSA/SHA1 signature": leapp_rpms_with_rsa_sha1_detected
   "The installed KDE environment is unavailable on RHEL 8": leapp_unavailable_kde
   "Cannot perform the VDO check of block devices": leapp_vdo_check_needed
+  # These titles have been replaced by a single title and should be removed with next release of leapp (Nov 2026)
```
pirat89:
```suggestion
  # These titles have been replaced by a single title and should be removed with next release of leapp-repository (Nov 2026)
```

do not mix leapp & leapp-repository products :) also keep in mind that they cannot be removed from here any time soon, as the systemrole is supposed to support also upgrade paths that do not receive the change (but stay still supported). IOW, it could be potentially removed in the playbook in 2y from now (2028). But the comment is written in a way it suggests it could be removed sooner.

roles/analysis/vars/main.yml
pirat89: see my comment above. it cannot be removed before 2028

roles/remediate/tasks/leapp_loaded_removed_kernel_drivers.yml
pirat89: again

roles/analysis/vars/main.yml
PeterMocary: I see, you are right. I'll adjust/remove the comments.

PeterMocary:
Addressed all the issues and added changelog fragment. The comments were removed as the removal can't be done before 2028, i don't see them as important as before.

richm:
lgtm - ready to merge?

PeterMocary:
[citest]

PeterMocary:
> lgtm - ready to merge?

I'll run the tests and we can merge.

spetrosi (approved):

PR #393, opened 2026-05-05
ci: Fix release workflow

spetrosi:
The --attach flag is not supported in the version of gh CLI available in GitHub Actions runners. Pass the tarball as a positional argument instead.
The checkout action and create-pull-request both configure auth, causing duplicate Authorization headers (HTTP 400 error).


spetrosi:
Tested in fork at https://github.com/spetrosi/infra.leapp/releases/tag/1.7.5

PR #406, opened 2026-07-13
Use GH feature to automatically generate relnotes for GH release

spetrosi:
Using CHANGELOG.rst for release notes is cumbersome because we have to select lines related to the current release. Otherwise the whole changelog is printed in each release RNs.
GH has a feature to automatically generate RNs that this repository had been using before. Let's stick to this to comply with GH practices. CHANGELOG.rst is still available in a file, no need to repeat it in GH RNs.

### DIFF
--- a/.github/workflows/release.yml
+++ b/.github/workflows/release.yml
@@ -68,7 +68,7 @@ jobs:
           git push origin "refs/tags/${VERSION}"
           gh release create "${VERSION}" \
             --title "infra.leapp ${VERSION}" \
-            --notes-file CHANGELOG.rst \
+            --generate-notes \
             --verify-tag \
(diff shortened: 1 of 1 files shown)


richm:
@spetrosi what do the generated notes look like compared to the old method?

spetrosi:
> @spetrosi what do the generated notes look like compared to the old method?

Exactly the same as this is the previously used method. It's the same as clicking the `Generate release notes` button in https://github.com/redhat-cop/infra.leapp/releases/edit/1.7.5

And doing `--notes-file CHANGELOG.rst \` pasted the whole changelog into release notes, which is not what we want.

richm (approved):

PR #407, opened 2026-07-28
feat: introduce leapp_secure_logging variable

spetrosi:
Replace hardcoded no_log: true with parametrized
no_log: "{{ leapp_secure_logging }}" so users can set leapp_secure_logging: false for debugging while maintaining secure defaults.


spetrosi:
[citest]

richm (approved):

PR #432, opened 2026-09-08
fix(common): append to leapp log instead of overwriting

prjagtap-rh:
## What does this PR do?
Fixes a bug in init_leapp_log where /var/log/leapp/ansible_leapp_* logs were overwritten during initialization.

## Changes

- Fixes log overwrites caused by using copy in a loop for both fact files and log sections.
- Writes fact files only to their dedicated paths:
    - /etc/ansible/facts.d/pre_ipu.fact
    - /etc/ansible/facts.d/non_rhel_packages.fact
- Records the job start time once in leapp_start_time to keep the log header consistent.
- Assembles the complete ansible_leapp_* log after all initialization data is collected, preserving:
    - Job header
    - ansible_facts
    - Non-RHEL package information
- Adds a role test to verify that all log sections are retained in the correct order.
- Adds a TMT plan to run the regression test.

Jira: RHEL-165208

## Summary by CodeRabbit

- **Bug Fixes**
  - Fixed Leapp log initialization to preserve complete job headers, collected facts, non-RHEL package information, and execution metadata.
  - Log sections are now retained in the correct order.
  - Fact files continue to be created independently alongside the complete log record.
  - Existing log files remain unchanged when initialization cannot proceed.

- **Tests**
  - Added automated coverage for log completeness, section ordering, fact-file creation, metadata preservation, failure handling, and cleanup behavior.


roles/common/tests/tests_init_leapp_log.yml
```diff
@@ -0,0 +1,117 @@
...
+            src: "{{ leapp_log_file }}"
+          register: log_content_slurp
+
+        - name: Test | Decode log content
+          ansible.builtin.set_fact:
+            log_content: "{{ log_content_slurp.content | b64decode }}"
+
+        - name: Test | Split log into sections
+          ansible.builtin.set_fact:
+            log_sections: "{{ log_content.strip().split('\n\n') }}"
```
PeterMocary: Nitpick: I'd prefer having every section named something and identifying it by the name. Not a blocker though, just a suggestion for next time.

roles/common/tasks/init_leapp_log.yml
```diff
@@ -30,12 +30,15 @@
       rename the log file to clear this failure and try again.
   when: log_file_stat.stat.exists
 
-# Don't forget to archive the log file at the end of the parent role
+- name: init_leapp_log | Record start time
+  ansible.builtin.set_fact:
+    leapp_start_time: "{{ now(fmt='%Y-%m-%dT%H:%M:%SZ', utc=true) }}"
+
 - name: init_leapp_log | Create new log file
   ansible.builtin.copy:
```
spetrosi, line 66: I guess the bug here is that the task uses the copy module that overwrites the file instead of amend?
You can amend with the following:
```
- name: Amend at the end of file
  ansible.builtin.lineinfile:
    insertafter: EOF
    path: ...
    line: ...
...
```

If you use lineinfile to amend you can keep the loops.

PeterMocary: That is correct, but the problem with copy was using it multiple times in a loop on one file. Currently, the log wont get overwritten because there is only one copy that initiates the whole file at once. I'm okay with the way it is now, although it could be cleaned up a bit.

The idea is, that we gather the data we need, store it in facts and only after everything is known we initiate the log file with single copy. The other places that add to this file either use command line append (`>>`) or the `lineinfile` you suggested. This yml file is initiation of the log file and thus it should overwrite if called multiple times. However, it used to overwrite its own entries which is now fixed by this PR.

PeterMocary: That is correct, but the problem with copy was using it multiple times in a loop on one file. Currently, the log wont get overwritten because there is only one copy that initiates the whole file at once. I'm okay with the way it is now, although it could be cleaned up a bit.

The idea is, that we gather the data we need, store it in facts and only after everything is known we initiate the log file with single copy. The other places that add to this file either use command line append (>>) or the lineinfile you suggested. This yml file is initiation of the log file and thus it should overwrite if called multiple times. However, it used to overwrite its own entries which is now fixed by this PR.

PeterMocary: You are correct - there is overwriting going on and we need to append instead. However, this file simply initializes the starting part of the log file (it has init in its name) - so this is where the file is created basically. We want to add multiple things into the file within this initialization (timestamp, some ansible facts, and so on). 

The problem here was that we gathered one part of the data and added it to a facts file and to the log file, then again gathered another part and used copy to new facts file and to the same log file. This way, we basically overwrite the file during the initialization. 

This PR simply gathers the data, adds them to the facts files and after it has everything it needs for the log initialization it then does the copy only once. Which is what we need, because every time this initialization is called we actually want the file to be overwritten, however, it is no longer being overwritten during the initialization. 

Just to note, other places that add new data into the log file after it is initialized are using append (either command line `>>` or the `lineinfile` you suggested).

That said, this approach is functional and IMO correct, although it could be cleaned up a bit.

roles/common/tasks/init_leapp_log.yml
```diff
@@ -78,15 +73,29 @@
...
+- name: init_leapp_log | Create new log file with complete initialization details
+  ansible.builtin.copy:
+    content: |
+      {{ job_name }}
+      Job started at {{ leapp_start_time }}
+
+      {{ ansible_facts | ansible.builtin.combine({'ansible_local': {}}) }}
+
+      {{ non_rhel_packages }}
+    dest: "{{ leapp_log_file }}"
```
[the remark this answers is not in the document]
PeterMocary: I made the check and creation of the file as close as possible to not increase the problem further. However, introduction of an atomic lock is outside of the scope of this PR.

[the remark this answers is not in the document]
PeterMocary: Yes, open an issue for this and make a detailed description with examples.

PeterMocary:
I restructured the original file to make the file creation as close to the check of its existence so that the lock is functional. There is a separate issue to make it atomic as this was not introduced in this PR. Other than that I added some internal variables so that the tests can test it with temporary files and packages. The test now tests 3 separate scenarios to increase the coverage.

roles/common/tests/tests_init_leapp_log.yml
```diff
@@ -0,0 +1,164 @@
...
+        - name: Test | Attempt init_leapp_log (expected to fail on existing log)
+          block:
+            - name: Test | Include init_leapp_log
+              ansible.builtin.include_role:
+                name: infra.leapp.common
+                tasks_from: init_leapp_log.yml
+          rescue:
+            - name: Test | Record that init_leapp_log failed as expected
+              ansible.builtin.set_fact:
+                init_failed: true
```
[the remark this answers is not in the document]
PeterMocary: This should be addressed already, resolving.

PeterMocary:
Rebased

PeterMocary:
Adjusted the tmt files to the new location.

PR #434, opened 2026-09-11
Restructure tests

mmacura311:
This PR moves tmt related code into separate directory.

## Summary by CodeRabbit

- **Tests**
  - Improved control-node setup reliability by recording Ansible collection requirement installation results and applying a 10-minute timeout.
  - Streamlined custom repository configuration during test preparation, including support for RHEL 7, RHEL 8+, and fallback repository settings.

- **Chores**
  - Added supporting test infrastructure to maintain the test library directory and manage repository configuration consistently.

PR #436, opened 2026-09-17
ci: Install collections using lsrAnsibleGalaxy that sets timeout to 300s

spetrosi:
Ansible Galaxy fails due to timeout, probably issues on Ansible server side.
This change runs the ansible-galaxy using lsrAnsibleGalaxy wrapper from https://github.com/linux-system-roles/tft-tests/pull/107 that patches ansible-galaxy of any version to set timeout of 300s

## Summary by CodeRabbit

* **Bug Fixes**
  * Dependency installation now reports detailed output and stops immediately when installation fails, providing clearer feedback during setup.


spetrosi:
Provided simplified version in https://github.com/redhat-cop/infra.leapp/pull/434#discussion_r4035886964
infra.leapp supports running only on newer ansible-core that has builtin `--timeout` parameter.

